PR #15, opened 2022-10-28
fix: fix handling of different object types passed into the SetOrPanic

nstott:
This fixes the handling of `SetOrPanic` so that it handles different object types more gracefully.

SetOrPanic will now

1. if an object is a reference, it will dereference that object 
  - this means we can pass in either *object or object directly, without having to worry about it
2. check to see if the object is a stringer, and if so it will string()ify the object.  
  - this means that we can serialize 'PointInTime' structs into a single string
 3. if the object is a slice or array, then it will serialize each element, and build a `[]interface{}`
  - this means we can pass in objects like  the ip allow lists, and set the entire slice in one go
 4. if the object is a struct, then it will wrap the struct in an `[]interface{}` and set that 

more comments inline


serdardalgic (approved):
Having two models `NewClusterForCreate` and `NewClusterForUpdate` looks ugly, IMHO. We might confuse them in the future, during any changes to this part of the code. However, I don't have a suggestion to get rid of this duality, and this structure works. 

But it's a fair enough trade-off for getting rid of `utils.NewPropList` calls and having a cleaner code. So, I approve this PR. There is one minor suggestion about the comments.

PR #21, opened 2022-10-31
check for cluster phase in our 'is cluster ok' check

### LINKED ISSUE #20: the check for completion should also check the cluster's `phase`
We currently do not check the cluster's `phase` when discovering if a change was successful.
we should wait until the cluster is in a healthy phase, as well as has a `biganimal.com/deployed` condition before we mark a change as being successful

nstott:
fixes #20 

Moves the `IsOk()` check from the api client to the cluster model, and also includes a check for a 'healthy' phase

merge after #19

### DIFF
--- a/pkg/models/cluster.go
+++ b/pkg/models/cluster.go
@@ -5,6 +5,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+const (
+	// from https://github.com/EnterpriseDB/dataplane-postgres-operator/blob/83d333f137280e8c1c7b27470f469aa58b0f763e/api/v1beta1/dataplanepostgrescluster_types.go#L301
+	CONDITION_DEPLOYED = "biganimal.com/deployed"
+
+	// from https://github.com/EnterpriseDB/upm-ui/blob/8840e856aa93f7c266196dc97a53da20d7578fe2/app/src/locales/en-US/global.ts#L69
+	PHASE_HEALTHY = "Cluster in healthy state"
+)
+
 func NewCluster(d *schema.ResourceData) (*Cluster, error) {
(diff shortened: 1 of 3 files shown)


pkg/models/cluster.go
```diff
@@ -5,6 +5,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+const (
+	// from https://github.com/EnterpriseDB/dataplane-postgres-operator/blob/83d333f137280e8c1c7b27470f469aa58b0f763e/api/v1beta1/dataplanepostgrescluster_types.go#L301
```
serdardalgic, line 9: @nstott do you want to reference to a private repository here? Also, in the other comments, I'm not pointing all.

pkg/provider/resource_cluster.go
```diff
@@ -361,8 +361,8 @@ func (c *ClusterResource) retryFunc(ctx context.Context, d *schema.ResourceData,
 			return resource.NonRetryableError(fmt.Errorf("Error describing instance: %s", err))
 		}
 
-		if !client.HasOkCondition(cluster.Conditions) {
-			return resource.RetryableError(errors.New("Instance not yet ready"))
+		if !cluster.IsHealthy() {
+			return resource.RetryableError(errors.New("Instance not yet health"))
```
serdardalgic, line 365:
```suggestion
			return resource.RetryableError(errors.New("Instance not yet ready"))
```
I'd reject this change and keep it with `Instance not yet ready` for now.

serdardalgic (changes requested):
I think we shouldn't reference to private repository codes/structures, even if they are in the comments. 

Let's try to find a way where we can get these information from [BA API](https://portal.biganimal.com/api/docs/) Docs e.g. https://portal.biganimal.com/api/docs/#/Clusters/getClusterPhases

PR #27, opened 2022-11-07
fix: export BA_API_URI variable

kunliuedb:
Provider use BA_API_URI to talk with BigAnimal API. Add the configuration step in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ export BA_BEARER_TOKEN=<REDACTED>
 
 After compiling, configuring the `.terraformrc` and fetching a token, the examples in the `./examples` folder can be run.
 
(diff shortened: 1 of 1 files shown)


serdardalgic (approved):
Good catch @kunliuedb , thank you! 
We have the BA_API_URI in the [.envrc file](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/.envrc#L1), @nstott uses that, and I was exporting this variable in my shell. That's how we forgot to add this information, I guess :)

nstott:
Good catch!

additionally, `https://portal.biganimal.com/api/v2` should become a default.  and then we can allow the override of that default with an env var

PR #31, opened 2022-11-07
Set OS_ARCH in GNUmakefile

### LINKED ISSUE #28: set OS_ARCH in GNUmakefile
`OS_ARCH` was set to `darwin_amd64` in [GNUmakefile](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/ba24021fd04942f80483bf8fe51e3f30132ed7d8/GNUmakefile#L8). It's not convenience for other OS. Open an issue to track.

serdardalgic:
fixes #28. 

That would be the solution I would propose.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -18,6 +18,9 @@ endif
 ifeq ($(shell uname -m),aarch64)
   arch   ?= arm
 endif
+ifeq ($(shell uname -m),arm64)
+  arch   ?= arm64
+endif
 OS_ARCH=${os}_${arch}
 
 default: install
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -5,7 +5,20 @@ NAMESPACE=edu
 NAME=biganimal
 BINARY=terraform-provider-${NAME}
 VERSION=0.3.1
-OS_ARCH=darwin_amd64
+
+# Figure out the OS and ARCH of the
+# builder machine
+os       ?= $(shell uname|tr A-Z a-z)
+ifeq ($(shell uname -m),x86_64)
+  arch   ?= amd64
+endif
+ifeq ($(shell uname -m),i686)
+  arch   ?= 386
+endif
+ifeq ($(shell uname -m),aarch64)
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -5,7 +5,20 @@ NAMESPACE=edu
...
+# Figure out the OS and ARCH of the
+# builder machine
+os       ?= $(shell uname|tr A-Z a-z)
+ifeq ($(shell uname -m),x86_64)
+  arch   ?= amd64
+endif
+ifeq ($(shell uname -m),i686)
+  arch   ?= 386
+endif
+ifeq ($(shell uname -m),aarch64)
```
kunliuedb, line 18: Hi Serdar, thank you for fixing this. I did a test on my M1 MacBook.  

```
kunliu@LAPTOP678-MA-US 0.3.1 % uname -m
arm64
```

Perhaps we can add one more IF condition here ?
```
ifeq ($(shell uname -m),arm64)
  arch   ?= arm64
endif
```

serdardalgic: That's a very good point, @kunliuedb, thanks for pointing out! 
The funny thing is, both `arm64` and `aarch64` technically should be the same thing, more info [here](https://en.wikipedia.org/wiki/AArch64). 
And our release workflow [generates](https://github.com/EnterpriseDB/terraform-provider-biganimal/releases/tag/v0.0.0-release) `arm64` executables, instead of `aarch64`. 

So, maybe we may not need `aarch64` at all, but I'm not sure. That's why, I'm only going to add `arm64` and not delete `aarch64`.

serdardalgic:
For the record, I've set a new tag for testing this PR
https://github.com/EnterpriseDB/terraform-provider-biganimal/releases/tag/v0.0.0-test

It doesn't break our go release workflow. @nstott, can you please review? Thank you.

mbiagetti (approved):
tested `make build` and works as expected (mac `intel i9`)

Issue #18, opened 2022-10-31
cluster_architecture is not sent in on resource updates


nstott:
We currently do not send the cluster_architecture in with an update because the api validation fails.
This is because of an un-set field within the ClusterArchitecture struct.

we should remove that field and re-enable the updating of this property

PR #33, opened 2022-11-07
fix: send the cluster_architecture in on cluster updates

nstott:
fixes https://github.com/EnterpriseDB/terraform-provider-biganimal/issues/18

This re-enables the cluster_architecture property and removes the 'name' field from the model to allow the validations to proceed

merge after https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/15

This PR is actually a clone of https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/19 . I opened a new one, because i mucked up the git history in that branch while trying to sign the commits :)

### DIFF
--- a/pkg/models/architecture.go
+++ b/pkg/models/architecture.go
@@ -1,7 +1,6 @@
 package models
 
 type Architecture struct {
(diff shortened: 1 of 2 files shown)


serdardalgic (approved):
I don't know how to test this change. 
But it LGTM.

PR #37, opened 2022-11-08
(chore): Update docs and examples via running make docs command

serdardalgic:


examples/cluster_resource/resource.tf
```diff
@@ -37,7 +37,7 @@ resource "biganimal_cluster" "this_resource" {
   }
 
   instance_type = "azure:Standard_D2s_v3"
-  password         = resource.random_password.password.result
+  password      = resource.random_password.password.result
```
drsm79, line 40: Can we do this with a linter? Is aligning the operator idiomatic to TF?

docs/data-sources/region.md
```diff
@@ -0,0 +1,42 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
```
drsm79, line 2: Be good to wire up https://terraform-docs.io/how-to/github-action/

nstott:
https://pre-commit.com/  is a good option to automate the doc build locally
see https://github.com/EnterpriseDB/upm-conventional-pre-commit for a specific example

docs/data-sources/region.md
serdardalgic: I've added this into #42 as well.

nstott (approved):
this looks good, as far as it goes, thanks

before we publish the provider, I think we should
- automate the doc generation
- linters
- check the doc text associated with the different properties to catch typos, bad copy pasta, etc
- add examples

PR #39, opened 2022-11-09
fix: change cloud_provider and region_id as the variables

kunliuedb:
A small change to pass cloud_provider and region_id as the variables when creating region resource.


serdardalgic (approved):
LGTM, thank you!

kunliuedb:
will create another PR which easy to sign the commits

Issue #42, opened 2022-11-09
Use Linter to check the terraform code and update the docs


serdardalgic:
We need to set up a mechanism to `terraform fmt` the code, and also `go generate` the docs and the examples. 
We can set up a pre-commit hook so that the developers/contributers can check their code. 

Even better, we can automate this system, such as https://terraform-docs.io/how-to/github-action/

serdardalgic:
Copying [this comment](https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/37#pullrequestreview-1174198856) @nstott here:

> before we publish the provider, I think we should
> * automate the doc generation
> * linters
> * check the doc text associated with the different properties to catch typos, bad copy pasta, etc
> * add examples

serdardalgic:
I think we can group the linters in such categories:
* Automated doc generation and terraform fmt check (can be addressed via https://terraform-docs.io/how-to/github-action/ ) 
* A conventional commit check. (e.g. https://github.com/marketplace/actions/conventional-commit-checker) 
* Non-inclusive and intrusive language check and typo check. 
* Naming convention check, to enforce the naming convention mentioned in the [CONTRIBUTING.md](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/CONTRIBUTING.md) file. 

For the naming and best-practice check, We can check https://github.com/terraform-linters/tflint.

PR #50, opened 2022-11-14
ci: Spellcheck GH action

serdardalgic:
This PR adds two checkers:
- spellchecker
- checker for non-inclusive language. 

The `.wordlist-en-custom.txt` file is generated after the output of the spellchecker. 

Please review. 

Related to #42.  Please merge it after merging #49

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -1,0 +1,13 @@
+BigAnimal
+CIDR
+GUC
+IOPS
+InstanceType
+Postgres
+biganimal
+config
+ip
+nestedatt
+nestedblock
+terraform
+uri
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -1,0 +1,31 @@
+name: spellcheck-woke
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+
+  # Check code for non-inclusive language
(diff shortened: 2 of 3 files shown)


drsm79 (approved):

.wordlist-en-custom.txt
```diff
@@ -0,0 +1,13 @@
+BigAnimal
+CIDR
+GUC
+IOPS
+InstanceType
+Postgres
+biganimal
```
nstott, line 7: should Biganimal be capitalized?

nstott (dismissed):
lgtm

.wordlist-en-custom.txt
serdardalgic: Unfortunately, spellchecker is not clever enough to distinguish when a word appears. This `biganimal` is coming from 
```
terraform {
  required_providers {
    biganimal = {
      source  = "biganimal"
      version = "0.3.1"
    }
  }
}
```

And, with the current implementation, if we use `biganimal` in, let's say, README.md, it will accept it. 🤷 not the best implementation, but I don't have any suggestions for fixing it. If you have, I'd love to hear. 

And I've only added the words that the spellchecker was complaining about. `Biganimal` wasn't one of them.

mbiagetti (approved):
LGTM

PR #51, opened 2022-11-14
Fix json pre commit errors inside the repo

serdardalgic:
This PR fixes the pre-commit checker errors for json files. 

The launch.json file is an autogenerated vscode-go file. VSCode uses Json with Comments(jsonc) mode for its configuration files.

If we can find a way to introduce jsonc check to our pre-commit hooks, we can change back the format of this file.

To read more about JsonC:
https://code.visualstudio.com/docs/languages/json#_json-with-comments

There is also an extension to JSON called Json5, Json for Humans. https://json5.org/ However, according to their website:
> it's not intended to be used for machine-to-machine communication. 

Please review.


drsm79 (approved):

serdardalgic:
I'm closing this PR in favor of #79. 
I think the JSON style changes doesn't bring any practical value for now. 
The indent change is trivial and not urgent. It can be tackled in another issue.

PR #68, opened 2022-11-17
ci: Configure CI pre-commit check

### LINKED ISSUE #53: Add a GH workflow to run precommit check
A pre-commit check file has been added within #49. 

It would be a nice to be able to run a GH workflow showing pre-commit checks on the PRs. Even if it's not a blocking PR check, it would at least report, and we can gradually improve the workflow. 

There is https://github.com/pre-commit/action, which is deprecated. They're suggesting https://pre-commit.ci/, but there are security concerns about giving a 3rd party GH app to be able to write to the PRs. It would be worth to run our own GH actions for pre-commit check.

mbiagetti:
Configure the ci to run pre-commit check at each commit/pr

Will scan only newer files compared to the main branch

close #53

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,29 @@
+name: Pre-Commit
+
+on:
+  workflow_dispatch:
+  push:
+
+env:
+  GOLANGCI_LINT_VERSION: "v1.50.1"
+
+jobs:
+
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: '0'
+
+      - name: Install required tools
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $HOME/.local/bin ${{ env.GOLANGCI_LINT_VERSION }}
+          golangci-lint --version
+
+      - name: Enforce pre-commit checks
+        uses: pre-commit/action@v3.0.0
+        with:
+          extra_args: --hook-stage push --from-ref origin/main --to-ref ${{ github.ref }}


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,29 @@
+name: Pre-Commit
+
+on:
+  workflow_dispatch:
+  push:
+
+env:
+  GOLANGCI_LINT_VERSION: "v1.50.1"
```
serdardalgic, line 8: I'm thinking if there would be a way to get this golanci-lint version from somewhere, but it looks complicated. 

[Our precommit hook](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/.pre-commit-config.yaml#L45) uses [the pre-commit-golang repo](https://github.com/dnephin/pre-commit-golang) and we do not describe the golangci-lint version there. AFAICS, pre-commit-golang repo also doesn't define it anywhere. So, it's fine as long as we have golangci-lint installed in the path. 

I wish the pre-commit/action took care of installation of the required tools though 🤷 but not possible for now :)

serdardalgic (approved):
Great Job! Thanks for the PR @mbiagetti!

As can be seen in this PR's checks, the pre-commit check works. LGtM!

.github/workflows/pre-commit.yml
mbiagetti: @serdardalgic exactly: as described in the [doc](https://github.com/dnephin/pre-commit-golang#available-hooks), we should take care of installing it.

Maybe we could use `latest` as a version, but usually I prefer to pin all the tools involved in our chain (like we already do in the [.pre-commit-config.yaml](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/.pre-commit-config.yaml) file, see the `rev` field)

serdardalgic: I strongly agree with you @mbiagetti, pinning the version sounds like a sane approach. Thanks for the links!

serdardalgic (approved):
LGTM

PR #75, opened 2022-11-22
fix: Increase Delete Timeout for region resource to 60 min

pizzacapricciosa:
Increased the delete timeout to 60 minutes for consistency


serdardalgic:
@pizzacapricciosa can you amend your commit to comply with [the conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), please? Thank you. 

A suggestion for the commit message would be :
```
fix: Increase Delete Timeout for region resource to 60 min
```

kunliuedb:
Create another PR #83 for the fix

PR #77, opened 2022-11-22
style: Fix some issues that pre-commit hook is complaining about

serdardalgic:
Also, let's exclude tools from go-vet for now. It throws the
```
build constraints exclude all Go files
```
error, but we don't know for now where it's coming from. 
<s>I'll create a GH issue for that</s>. #78 is created to track that problem


mbiagetti (approved):
LGTM 👍

PR #84, opened 2022-11-22
fix: Set ba_bearer_token and ba_api_uri provider arguments

serdardalgic:
With this PR, BigAnimal provider block can take arguments. Before, terraform provider wasn't configurable via provider block. 
```hcl
provider "biganimal" {
  ba_bearer_token = <redacted> // See Getting and API Token section for details
  // ba_api_uri   = "https://portal.biganimal.com/api/v2" // Optional
}
```
* The naming for the provider block arguments and the environment variables are now the same.(before it was `ba_token` for argument and `BA_BEARER_TOKEN` for the env var.  
* I tried to update the README.md, I'm open to suggestions.
* ba_api_uri argument is added as optional, with the default value `"https://portal.biganimal.com/api/v2"`. 
* I would update the examples to include a provider configuration in a separate PR. 

Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The binary can also be compiled by `go build`, which will output the binary into
 
 ## Using the provider
 
-Until the provider is accepted into the terraform registry, it's necesary to install the binary into your local terraform cache using `make`, and to configure terraform to look in the right location to find the binary.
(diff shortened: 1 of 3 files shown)


nstott (approved):
love it, thanks!

PR #87, opened 2022-11-23
Add EHA resource example and remove deprecated replicas

kunliuedb:
This PR address two issues:

- Add resource example for EHA cluster
- Remove deprecated `replicas`


serdardalgic (approved):
LGTM, great job @kunliuedb, thank you very much 🙏

PR #90, opened 2022-11-29
docs: Correct the token variable name

serdardalgic:
We standardized the provider block parameter `ba_bearer_token` and the env var name `BA_BEARER_TOKEN` in #84 , but 
I forgot to update the examples. 

This PR fixes the example and the documentation. 

Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ provider_installation {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


kunliuedb (approved):
LGTM

PR #91, opened 2022-11-29
ci: Prepare for the release

serdardalgic:
* Adjust the HOSTNAME variable in GNUmakefile
* Remove the NAMESPACE, not needed
* Update the version to 0.1.0
* Update the source string in docs and examples

To test this change, please ensure that your `~/.terraformrc` file is similar to this:
```
provider_installation {

  direct {
    exclude = ["registry.terraform.io/EnterpriseDB/biganimal"]
  }
  filesystem_mirror {
    path    = "/Users/<username>/.terraform.d/plugins"
    include = ["EnterpriseDB/biganimal"]
  }
}
```
This change will be merged later with #89. This terraformrc hack is needed until we publish the provider to the terraform registry.

PR #93, opened 2022-11-29
ci: Correct the hostname and namespace settings in the Makefile

serdardalgic:
I removed the namespace variable and set the hostname to an invalid value in #91. 
This PR addresses the issue.


nstott (approved):
lgtm
thanks!

Issue #47, opened 2022-11-10
Include version to the UA header to identify the provider and its version


drsm79:
We should have a/some HTTP header(s) which:

- clearly identifies the request as coming from the provider
- records version of teraform
- records version of provider

I think this should be [added here](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/pkg/api/api.go#L17)

drsm79:
The user-agent identifies the request as coming from the provider, but not versions.

serdardalgic:
Actually, the headers are added when the request is being prepared. That's why the user agent is set during the doRequest function here https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/pkg/api/request.go#L19.

The constant that's set is defined inside the [cluster_client.go](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/pkg/api/cluster_client.go#L14) 

I don't know how, but we can get [this version string](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/main.go#L20) (this can also be set by goreleaser while we're preparing the release packages) and we can add this version string to the end of the constant) 

@nstott any thoughts on that?

PR #97, opened 2022-12-01
fix: Change api code structure for setting user-agent

YanniHu1996:
to add `user-agent` to the HTTP header, make the following changes

- change `NewXxxClient(ba_bearer_token string, ba_api_uri string)` to ``NewXxxClient(api API)`
- change `doRequest()` called in each methods to `api.doReuqest()`

methods in `XxxClient` don't need to know `token` and `url`

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -7,33 +7,34 @@ import (
 )
 
 type API struct {
-	Url   string
-	Token string
-
-	httpClient *http.Client
+	BaseURL    string
+	Token      string
+	UserAgent  string
+	HTTPClient http.Client
 }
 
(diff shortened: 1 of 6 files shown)


serdardalgic (approved):
I've played a bit with the PR, I was thinking of changing [the VERSION string in GNUmakefile](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/GNUmakefile#L7) btw, but then I reverted those changes. That's the reason of my pushed commits in this PR :) 

I've created a new cluster with the code in this PR successfully. I was able to track the User-Agent string `terraform-provider-biganimal/dev` and it works as expected. 

I'm approving this PR and merging it to the main. 

fixes #47

PR #100, opened 2022-12-06
Better Error Handling 

nstott:


serdardalgic:
So, here is my current implementation:
Screenshot 2022-12-07 at 16 36 14

It prints the error message, and all the details are provided in the description section. 

I think this PR is ready for review. @nstott I'm rewording your first commit to `fix: Initial commit for better error handling` and rebasing the PR. (otherwise, the conventional-commit checker complains)

examples/resources/biganimal_cluster/single_node/resource.tf
```diff
@@ -63,7 +63,7 @@ resource "biganimal_cluster" "single_node_cluster" {
   pg_version            = "14"
   private_networking    = false
   cloud_provider        = "azure"
-  read_only_connections = false
+  read_only_connections = true
```
serdardalgic, line 66: This change is for testing the API Errors. 
It must be removed before merging this PR.

pkg/api/errors.go
```diff
@@ -1,44 +1,75 @@
...
+
+	apiErr := APIError{}
+
+	if json.Unmarshal(body, &apiErr) != nil {
+		return errors.New("unknown API Error")
+	}
+
+	return &BigAnimalError{
+		APIError: apiErr,
+		Err:      fmt.Errorf("status: %d - %v", apiErr.Error.Status, apiErr.Error.Message),
```
serdardalgic, line 73: There is a repetition with the Error string of the BigAnimalError above
```golang
func (baerr *BigAnimalError) Error() string {
	return fmt.Sprintf("status: %d - %v", baerr.APIError.Error.Status, baerr.APIError.Error.Message)
}
```
Is there a way to avoid this? 🤔

pkg/api/errors.go
```diff
@@ -1,44 +1,75 @@
...
+
+func (baerr *BigAnimalError) getDetails() string {
+	var details string
+	for _, err := range baerr.APIError.Error.Errors {
+		details += fmt.Sprintln(err.Message)
+	}
+	return details
+}
+
+func FromErr(err error) diag.Diagnostics {
```
serdardalgic, line 42: api.FromErr tries to assert the BigAnimalError type to the error. 
If this succeeds, it returns a crafted diag.Diagnostics 
If not, it calls the `diag.FromErr` function. 

I've changed the `diag.FromErr` to `api.FromErr` function only after the Client function calls, which call the `doRequest` function. 
Technically, any `diag.FromErr` function can be changed with `api.FromErr` for the sake of expediency. @nstott what do you think about that? I don't have a strong opinion on that.

pkg/utils/token.go
```diff
@@ -0,0 +1,6 @@
+package utils
```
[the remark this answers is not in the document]
nstott: yeah, this should be removed.  it was accidentally added when i made the PR
thanks!

pkg/api/cluster_client.go
```diff
@@ -79,7 +80,7 @@ func (c ClusterClient) ReadByName(ctx context.Context, name string) (*models.Clu
 	}
 
 	if len(clusters.Data) != 1 {
-		return &models.Cluster{}, Error404
+		return &models.Cluster{}, errors.New("resource Not Found")
```
nstott, line 83: this should be a static error.
if an error is static, then we can make tests and compare concrete error types
if we generate the error at runtime like this, then it's more awkward to test, and we have to copy strings as in the test below

pkg/api/errors.go
nstott: tbh, I think we should avoid embedding the diag's into the api package.

the library code should try to return standard golang error types.  `BigAnimalError` is a good way to do that

we can compare error types with error `Is()` and `As()` https://gosamples.dev/check-error-type/ https://pkg.go.dev/errors#As

errors.Wrap might also be useful if we are chaining errors

nstott (commented):
looks good!

two changes that I think should be addressed before merge
- make the 'not found error' a concrete error type.  define it once, rather than calling errors.New multiple times
- avoid embedding the terraform diags into the api package.  let's have the api package just return an 'error', and have the terraform bits in the terraform package only

pkg/api/errors.go
serdardalgic: 100% agree with keeping the api package clean, without the sdk/diag import. 
I've moved the FromErr function to `pkg/provider/utils.go` and named it `BigAnimalFromErr`.(any better naming suggestions are welcome, I couldn't come up with something better 😅)
I've written the comparison with `As()`.

pkg/api/errors.go
nstott: I think we can just have 
```
type BigAnimalError struct {
	APIError APIError
}

func (baerr *BigAnimalError) Error() string {
	return fmt.Sprintf("status: %d - %v", baerr.APIError.Error.Status, baerr.APIError.Error.Message)
}

```

and then when we use it in the diags, we can just call it 
```
		return diag.Diagnostics{
			diag.Diagnostic{
				Severity: diag.Error,
				Summary:  baError.Error(),
				Detail:   baError.GetDetails(),
			},
		}
```

pkg/provider/utils.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/EnterpriseDB/terraform-provider-biganimal/pkg/api"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+)
+
+func FromBigAnimalErr(err error) diag.Diagnostics {
```
nstott, line 12: we should make this lower case, so that we don't export it 
keep the interface small

nstott (commented):
nice!
couple of minor tweaks and it's good to go

pkg/api/errors.go
serdardalgic: Amazing! Thanks for the hint, @nstott! 🚀 🤘

pkg/provider/utils.go
serdardalgic: Ah, good catch! Done.

nstott (commented):
lgtm!

serdardalgic (approved):

PR #111, opened 2023-01-09
feat: Introduce logs_url and metrics_url to the biganimal_cluster resource/data-sources

serdardalgic:
With this change, the cluster data sources will be able to read the logs and metrics URLs.
Because they are computed fields, we're not reading values for those fields while creating a cluster. 

Please review.


nstott (approved):
lgtm! 
looks great

Issue #85, opened 2022-11-23
Add Validations for some parameters in the resources


serdardalgic:
While defining the Schema, we can use [ValidateDiagFunc](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatediagfunc) for the validation of the values for primitive types. 

Some examples would be:
* biganimal_cluster:
  * backup_retention_period
  * instance_type
  * password
  * pg_type
  * id field of cluster_architecture
  * region values for the cloud providers. For example, it's `eastus` and not `useast` for the Azure cloud provider.

serdardalgic:
We migrated most of our resources to the Terraform Framework Library, and the validations are a bit different then the SDKv2 way of doing things. 

There is a dedicated page for [the Validations in TF Framework](https://developer.hashicorp.com/terraform/plugin/framework/validation). 

We should also be careful about which values we validate. The validations that are already done on the API side should not be repeated in the TF Provider code, unless they're bringing a certain value to the customers. 

The CSP specific validations can be enhanced in the future. There are no CSP specific validations at this moment. 

#326 has been created to address current need for validations. When this ticket is merged, we can close this Issue as well.

PR #113, opened 2023-01-10
feat: Adapt to BigAnimal APIv3

serdardalgic:
The PR looks good to go. 

I've created single_node clusters on AWS and Azure, EHA and HA clusters on AWS, and they all looked fine. 
Regions look good to go as well. 

With this PR, we introduce the `validators.go`. It's quite straightforward for a single field validation, which can be used later to accomplish the #85. 

It also contains a fix about the logs_url and metrics_url for the clusters. 

Please review.

### DIFF
--- a/pkg/api/cluster_client.go
+++ b/pkg/api/cluster_client.go
@@ -23,7 +23,7 @@ func NewClusterClient(api API) *ClusterClient {
 	return &c
 }
 
-func (c ClusterClient) Create(ctx context.Context, model any) (string, error) {
+func (c ClusterClient) Create(ctx context.Context, projectId string, model any) (string, error) {
 	response := struct {
 		Data struct {
 			ClusterId string `json:"clusterId"`
@@ -37,7 +37,8 @@ func (c ClusterClient) Create(ctx context.Context, model any) (string, error) {
 		return "", err
 	}
 
-	body, err := c.doRequest(ctx, http.MethodPost, "clusters", bytes.NewBuffer(b))
+	url := fmt.Sprintf("projects/%s/clusters", projectId)
+	body, err := c.doRequest(ctx, http.MethodPost, url, bytes.NewBuffer(b))
 
 	if err != nil {
 		return "", err
@@ -47,12 +48,12 @@ func (c ClusterClient) Create(ctx context.Context, model any) (string, error) {
 	return response.Data.ClusterId, err
 }
 
-func (c ClusterClient) Read(ctx context.Context, id string) (*models.Cluster, error) {
+func (c ClusterClient) Read(ctx context.Context, projectId, id string) (*models.Cluster, error) {
 	response := struct {
 		Data models.Cluster `json:"data"`
 	}{}
 
-	url := fmt.Sprintf("clusters/%s", id)
+	url := fmt.Sprintf("projects/%s/clusters/%s", projectId, id)
 	body, err := c.doRequest(ctx, http.MethodGet, url, nil)
 	if err != nil {
 		return &response.Data, err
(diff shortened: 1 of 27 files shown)


pkg/api/cluster_client.go
```diff
@@ -63,12 +64,12 @@ func (c ClusterClient) Read(ctx context.Context, id string) (*models.Cluster, er
...
 }
 
-func (c ClusterClient) ReadByName(ctx context.Context, name string) (*models.Cluster, error) {
+func (c ClusterClient) ReadByName(ctx context.Context, projectId, name string) (*models.Cluster, error) {
 	clusters := struct {
 		Data []models.Cluster `json:"data"`
 	}{}
 
-	url := fmt.Sprintf("clusters?name=%s", name)
+	url := fmt.Sprintf("projects/%s/clusters?name=%s", projectId, name)
```
nstott, line 72: is it really this easy?

pkg/provider/data_source_cluster.go
```diff
@@ -165,6 +165,11 @@ func (c *ClusterData) Schema() *schema.Resource {
 				Type:        schema.TypeBool,
 				Computed:    true,
 			},
+			"project_id": {
+				Description: "BigAnimal Project ID.",
```
nstott, line 169: is it worth validating the input here and make sure that it's in the form /prj_.*/ ?

pkg/provider/data_source_cluster.go
```diff
@@ -237,16 +242,20 @@ func (c *ClusterData) Read(ctx context.Context, d *schema.ResourceData, meta any
 
 	clusterName, ok := d.Get("cluster_name").(string)
 	if !ok {
-		return diag.FromErr(errors.New("Unable to find cluster name"))
+		return diag.FromErr(errors.New("unable to find cluster name"))
+	}
+	projectId, ok := d.Get("project_id").(string)
+	if !ok {
+		return diag.FromErr(errors.New("unable to find a project ID"))
```
nstott, line 249: > unable to find a project ID

just a nit, but it might be more accurate to say 
> unable to find project with id '%s'

or something?

pkg/provider/resource_cluster.go
```diff
@@ -370,13 +402,17 @@ func (c *ClusterResource) Delete(ctx context.Context, d *schema.ResourceData, me
...
+		}
+		cluster, err := client.Read(ctx, projectId, clusterId)
 		if err != nil {
-			return resource.NonRetryableError(fmt.Errorf("Error describing instance: %s", err))
+			return resource.NonRetryableError(fmt.Errorf("error describing instance: %s", err))
 		}
 
 		if !cluster.IsHealthy() {
-			return resource.RetryableError(errors.New("Instance not yet ready"))
+			return resource.RetryableError(errors.New("instance not yet ready"))
```
nstott, line 415: thanks for the consistency!

pkg/api/cluster_client.go
serdardalgic: Yeah. Actually, the API v3 and API v2 outputs are the same, just the endpoints are changing. 
I didn't find anything else 🤷 . Anyhow, we need to test this branch thoroughly.

pkg/provider/data_source_cluster.go
serdardalgic: This actually checks the `project_id` value in the configuration. So, it doesn't check if a project with this `project_id` exists or not. 

I'm thinking about checking if the `project_id` project is reachable or not, even before the calls. I will check tomorrow.

pkg/provider/data_source_cluster.go
serdardalgic: Aaand, we have our first `validators.go` as well as `validators_test.go` 😄

pkg/provider/data_source_cluster.go
serdardalgic: This check is actually pretty useless. `project_id` is always a required field, It should always be provided by the user to run the terraform command on that resource or data-source. Only after that, the validation function will run and validate that the project_id starts with `prj_`, followed by 16 letters. So, this error check only shows if the 
 `d.Get("project_id").(string)` is failed. I think this is enough to discard this `if !ok {` check. 

We are already doing it in the region resource btw, for example https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/pkg/provider/resource_region.go#L104-L106

I think it's fair to remove them. What do you think, @nstott?

kunliuedb (approved):
LGTM

Issue #25, opened 2022-11-04
Resource Not Found returned if there are clusters with the same name


kunliuedb:
### Terraform Version
Terraform v1.3.2

### Affected Resource(s)

- cluster_data

### Actual Behavior
`Error: Resource Not Found
with data.biganimal_cluster.this,
on data.tf line 10, in data "biganimal_cluster" "this":
10: data "biganimal_cluster" "this" {
`

### Root cause
The root cause is the logic [here](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/072f3aa96819265ddd3ce57d02f350c7c3364a61/pkg/api/clusterClient.go#L97)
`if len(clusters.Data) != 1 {
		return &models.Cluster{}, Error404
}`

serdardalgic:
I would like to close this PR with the next release, when we release the solution implemented in #118.

PR #118, opened 2023-01-16
fix: Show most_recent cluster when there are multiple with the same name

### LINKED ISSUE #43: Update cluster data_source to handle multiple clusters with the same name
In BigAnimal, a cluster is identified by either [provider, region, name] or [clusterid], cluster's name itself won't be able to identify a cluster by itself.

In the current implementation, we use cluster names to fetch the clusters via data_sources. If there are multiple clusters with the same name, [we return 404](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/072f3aa96819265ddd3ce57d02f350c7c3364a61/pkg/api/clusterClient.go#L97). There is a bug reported by @kunliuedb at #25. 

I'd like to propose the following solution:
* `biganimal_cluster` data_source would get `cluster_name` as an argument and would fail if there are more than 1 clusters returned. The error message should be improved to inform the user that there are multiple clusters with the same name. 
* In the future, we can think of providing a solution like `the most recent` one. This can be a data_source argument to overcome the fail. 
* We can introduce a new data_source called `biganimal_cluster_ids` that would return the ids of the clusters with a given name
* We can include `cluster_ids` as an argument for the `biganimal_cluster` data source. This data_source would check if there is a name or cluster_id so that it can fetch the cluster information from the API, and return as a data_source. Name and cluster_id should be mutually exclusive and terraform should fail if they both exist in the data_source.

serdardalgic:
This fixes #43 and #25. 

For clarity, I sorted the `clusters.data` response according to the Seconds. For better granularity, Nanos check can also be included. 

`most_recent` field has a default value of `true`. 

Please review.

### DIFF
--- a/docs/data-sources/cluster.md
+++ b/docs/data-sources/cluster.md
@@ -113,7 +113,7 @@ output "storage" {
 
 ### Optional
 
-- `most_recent` (Boolean) Show the most recent cluster when there are multiple clusters with the same name. It defaults to true.
+- `most_recent` (Boolean) Show the most recent cluster when there are multiple clusters with the same name.
 
 ### Read-Only
 
--- a/pkg/provider/data_source_cluster.go
+++ b/pkg/provider/data_source_cluster.go
@@ -77,10 +77,10 @@ func (c *ClusterData) Schema() *schema.Resource {
 				Required:    true,
 			},
 			"most_recent": {
-				Description: "Show the most recent cluster when there are multiple clusters with the same name. It defaults to true.",
+				Description: "Show the most recent cluster when there are multiple clusters with the same name.",
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     true,
+				Default:     false,
 			},
 			"created_at": {
 				Description: "Cluster creation time.",


pkg/api/cluster_client.go
```diff
@@ -80,7 +81,11 @@ func (c ClusterClient) ReadByName(ctx context.Context, projectId, name string) (
 	}
 
 	if len(clusters.Data) != 1 {
-		return &models.Cluster{}, Error404
+		if most_recent {
+			sort.Slice(clusters.Data, func(i, j int) bool { return clusters.Data[i].CreatedAt.Seconds > clusters.Data[j].CreatedAt.Seconds })
+		} else {
+			return &models.Cluster{}, Error404
```
kunliuedb, line 87: Just suggest to return meaningful message here instead of `resource not found`. If it falls into this branch, it means that there is more than one resource with the same cluster name in the same project (which I admit it's rare).

serdardalgic: but it can happen, right? :) So, I'm updating the error message, thanks for the suggestion @kunliuedb 👍

serdardalgic: Wrote [here](https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/118#issuecomment-1385347383)

pkg/api/cluster_client.go
```diff
@@ -80,7 +81,11 @@ func (c ClusterClient) ReadByName(ctx context.Context, projectId, name string) (
 	}
 
 	if len(clusters.Data) != 1 {
-		return &models.Cluster{}, Error404
+		if most_recent {
+			sort.Slice(clusters.Data, func(i, j int) bool { return clusters.Data[i].CreatedAt.Seconds > clusters.Data[j].CreatedAt.Seconds })
+		} else {
+			return &models.Cluster{}, ErrorClustersSameName
```
dhilipkumars, line 87: should `most_recent` be a default option if multiple clusters found? 
may be print a log saying that
`we found multiple clusters but returned the most recent one?`

dhilipkumars (commented):
MinorComment

pkg/api/cluster_client.go
dhilipkumars: the reason im thinking is we made the api-call, we got the data and we are dropping it at the last min and asking the user to make another call?

kunliuedb (dismissed):
LGTM

pkg/api/cluster_client.go
serdardalgic: That's a good suggestion @dhilipkumars. 
In our current design, the client sends the return value and the errors. When we find two or more clusters with the same name, we can return this info from the client and turn it into a `Warning` level Diagnostic, but I need to think about how to do that. 

For now, the best I can offer is creating a ticket for this and adding a `tflog.Info` when most_recent is true.

serdardalgic: Actually, I'm not sure about the `tflog.Info` there because it's the client code. We're trying to keep the client code as terraform-agnostic as possible. So, I will write the debug info in the ticket as well.

serdardalgic: I created #121 so that we can tackle the problem in a dedicated time. 
Do you agree with that @dhilipkumars? If It's not a blocker, I'd like to merge this PR first, and then tackle the Warning system later.

dhilipkumars: Thanks @serdardalgic im resolving this comment

Issue #98, opened 2022-12-02
Repeated terraform apply reports "1 changed" in the output even though there are no changes in TF files.


ftouserkani-edb:
**% terraform -v**
```
Terraform v1.3.6
on darwin_amd64
```
**### Terraform Configuration Files**
```
terraform {
  required_providers {
    biganimal = {
      source  = "EnterpriseDB/biganimal"
      version = "0.1.0"
    }
    random = {
      source  = "hashicorp/random"
      version = "3.3.1"
    }
  }
}

resource "random_password" "password" {
  length           = 16
  special          = true
  override_special = "{}{}()"
  min_lower        = 3
  min_numeric      = 5
  min_special      = 3
  min_upper        = 2
}

resource "random_string" "test" {
  length = 20
  special          = true
  override_special = "/@£$?"
  min_lower        = 5
  min_numeric      = 5
  min_upper        = 5
  min_special      = 5
}

resource "biganimal_cluster" "single_node_cluster" {
  cluster_name = var.cluster_name

  allowed_ip_ranges {
    cidr_block  = "127.0.0.1/32"
    description = "localhost"
  }

  allowed_ip_ranges {
    cidr_block  = "192.168.0.1/32"
    description = "description!"
  }

  backup_retention_period = "6d"
  cluster_architecture {
    id    = "single"
    nodes = 1
  }

  instance_type = "azure:Standard_D2s_v3"
  password      = resource.random_password.password.result
  pg_config {
    name  = "application_name"
    value = "created through terraform"
  }

  pg_config {
    name  = "array_nulls"
    value = "off"
  }

  storage {
    volume_type       = "azurepremiumstorage"
    volume_properties = "P1"
    size              = "4 Gi"
  }

  pg_type               = "epas"
  pg_version            = "14"
  private_networking    = false
  cloud_provider        = "azure"
  read_only_connections = false
  region                = "eastus2"
}

provider "biganimal" {
  ba_bearer_token = "<redacted>"
}
```

### Expected Behavior
There shouldn't be any changes pending that needs to be applied since there hasn't been any TF file changes.
Please see the output regarding "resizing_vpc" and "storage iops and throughput". 

**### Actual Behavior**
**% terraform apply**
```
random_password.password: Refreshing state... [id=none]
random_string.test: Refreshing state... [id=3w8�WEJs0enw17E/�I?�]
[...]

Issue #117, opened 2023-01-16
Upgrade to terraform plugin framework. 


serdardalgic:
When we started writing the terraform provider, [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework) was under active development, so we built the provider using [the Terraform Plugin SDK v2](https://github.com/hashicorp/terraform-plugin-sdk). 

The [1.0.0 version](https://github.com/hashicorp/terraform-plugin-framework/releases/tag/v1.0.0) of plugin framework has been released on December, with semantic versioning compatibility promises. 

It would be a wise decision to switch to the plugin framework at some point, preferably before the SDKv2 support is ceased in the future. The steps are described in this [Migrate to Terraform Plugin Framework page](https://developer.hashicorp.com/terraform/plugin/sdkv2#migrate-to-terraform-plugin-framework).

Issue #41, opened 2022-11-09
Update the authentication to use refresh_token for a more maintainable authentication experience


serdardalgic:
Right now, we're using API Token as Bearer token, more details [here](https://github.com/EnterpriseDB/terraform-provider-biganimal#getting-an-api-token). API Token has to be renewed everyday. 

We're able to get the `refresh_token` within the API token fetching process. This refresh token lives for 30 days, and can be used to refresh the auth token. 

We can implement a mechanism that would use the refresh token instead of auth token (or both). The users would provide `refresh token` as a must, `auth token` as optional. When the auth token is expired, the refresh_token would be used to fetch a new auth token. If the refresh token doesn't work, the authentication fails.

PR #123, opened 2023-01-24
Terraform plan diff improvements

serdardalgic:
This PR addresses **some of** the problems mentioned in #98 
* ~~`allowed_ip_ranges` and `pg_config` will be [TypeSet](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typeset) because their order might change and that had generated unnecessary changes in the plan output~~ This has been moved to #128, see [the comment below](https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/123#issuecomment-1412116738) for the reasons. 
* storage throughput value in `biganimal_cluster` resource is now computed because it is automatically calculated by BigAnimal based on the IOPS input.
* DiffSuppressFunc is set for IOPS field of the storage. If there is no initial value, and it already has a set value in the state, it won't show any diff for this field. 

These changes require a Minor bump and they don't require any state migrations. 

Please review.

### DIFF
--- a/pkg/provider/resource_cluster.go
+++ b/pkg/provider/resource_cluster.go
@@ -36,7 +36,7 @@ func (c *ClusterResource) Schema() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"allowed_ip_ranges": {
 				Description: "Allowed IP ranges.",
-				Type:        schema.TypeSet,
+				Type:        schema.TypeList,
 				Optional:    true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


serdardalgic:
As stated in [the Example Major Number Increments section](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/versioning#example-major-number-increments) of the Versioning Best Practices document: 
> Changing attribute type where the new type is functionally incompatible (including but not limited to changing `TypeSet` to `TypeList` and `TypeList` to `TypeSet`)
would required a Major version bump.

That's why, I'd like to remove this change from this PR and add a separate PR for that change. 
We can include that change before a major bump. The next major bump will be our way to `1.0.0` which would (quoting the same document above)
> Version numbers above 1.0.0 signify stronger compatibility guarantees, based on the rules above.

Probably we might have some more groundbreaking changes before our 1.0.0 releases, to name a few
* #41 
* #117

pkg/provider/resource_cluster.go
```diff
@@ -226,16 +226,26 @@ func (c *ClusterResource) Schema() *schema.Resource {
...
 							Optional:    true,
+							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+								// iops is an optional field.
+								// If there is already a value set (old != "")
+								// and there is no new value (new == ""),
+								// we can suppress this Diff
+								if new == "" && old != "" {
+									return true
+								}
+								return false
```
mbiagetti, line 237: this could be simplified with 
```suggestion
								return new == "" && old != ""
```

mbiagetti (approved):
LGTM

PR #125, opened 2023-01-25
Implementation of Projects 

serdardalgic:
It's still a draft, but the `biganimal_projects` data-source work. 

There might be breaking changes still, so, play cautiously with it.


pkg/provider/resource_project.go
```diff
@@ -0,0 +1,173 @@
...
+func (p *ProjectResource) Update(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if d.HasChange("project_name") {
+		client := api.BuildAPI(meta).ProjectClient()
+		projectId := d.Id()
+		projectName := d.Get("project_name").(string)
+		_, err := client.Update(ctx, projectId, projectName)
+		if err != nil {
+			return fromBigAnimalErr(err)
+		}
+		err = resource.RetryContext(
```
YanniHu1996, line 143: `client.Update` is a synchronous invocation, we can know if the project is created successfully according to the return of this method. So it looks like we don't need to call `retryFunc` as well as `Create`

pkg/provider/data_source_projects.go
```diff
@@ -0,0 +1,100 @@
...
+	query := d.Get("query").(string)
+
+	projects, err := client.List(ctx, query)
+	if err != nil {
+		return fromBigAnimalErr(err)
+	}
+
+	utils.SetOrPanic(d, "projects", projects)
+	// FIXME: Check if there is a better way to set the ID.
+	d.SetId(strconv.FormatInt(time.Now().Unix(), 10))
```
YanniHu1996, line 97: how about `uuid`?

serdardalgic: Which `uuid` do you mean? Can you write as a Github code suggestion?

YanniHu1996 (approved):
LGTM

PR #145, opened 2023-02-28
feat: added support for faraway replica resource operations.

edbamolvyavahare:


.gitignore
```diff
@@ -27,6 +27,7 @@ terraform.tfvars
 *.iml
 .dccache
 website/vendor
+vendor/**
```
serdardalgic, line 30: We can delete this addition as #148 is merged.

examples/resources/biganimal_fa_replica/aws/resource.tf
```diff
@@ -0,0 +1,66 @@
...
+  type        = string
+  description = "BigAnimal source cluster ID"
+}
+
+variable "project_id" {
+  type        = string
+  description = "BigAnimal Project ID"
+}
+
+resource "biganimal_fareplica" "faraway_replica" {
```
serdardalgic, line 29: I think we should go with the explicit name of `biganimal_faraway_replica` as the resource name instead of `biganimal_fareplica`. WDYT? 

Also, the example directory name should be equivalent to the resource. For example, a `biganimal_faraway_replica` resource should have an example under `examples/resources/biganimal_faraway_replica/resource.tf` so that the [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) can add the code snippet to the corresponding documentation. 

The official documentation docs are [here](https://developer.hashicorp.com/terraform/registry/providers/docs).

pkg/provider/resource_cluster.go
```diff
@@ -264,6 +271,10 @@ func (c *ClusterResource) Schema() *schema.Resource {
 func (c *ClusterResource) Create(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := api.BuildAPI(meta).ClusterClient()
 
+	err := d.Set("cluster_type", "cluster")
+	if err != nil {
+		return nil
```
serdardalgic, line 276: If there is an error, why are we returning nil? 🤔

pkg/provider/resource_cluster.go
```diff
@@ -87,6 +87,13 @@ func (c *ClusterResource) Schema() *schema.Resource {
 					},
 				},
 			},
+
+			"cluster_type": {
```
serdardalgic, line 91: We need to write [a state migration](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/state-migration#terraform-v0-12-sdk-state-migrations) for adding a new field to the biganimal_cluster schema. 

You can test the state migration as follows:
* Create a biganimal_cluster using the latest terraform-provider-biganimal.
* Make sure you have the cluster information in the terraform state. You can run `terraform plan/apply` after a successful cluster creation, and it says there are no changes to apply. 
* After that, you compile your local code and start using the compiled terraform-provider-biganimal that includes this `cluster_type` change and try running `terraform plan` on the existing biganimal_cluster resource. 
* Because it will read the terraform state first, it won't find the `cluster_type` field in the `biganimal_cluster` schema. The state migration is necessary at that step.

pkg/provider/resource_fareplica.go
```diff
@@ -0,0 +1,369 @@
...
+
+type FAReplicaResource struct{}
+
+func NewFAReplicaResource() *FAReplicaResource {
+	return &FAReplicaResource{}
+}
+
+func (c *FAReplicaResource) Schema() *schema.Resource {
+	return &schema.Resource{
+		//todo: Description: "The cluster resource is used to manage BigAnimal clusters. See [Creating a cluster](https://www.enterprisedb.com/docs/biganimal/latest/getting_started/creating_a_cluster/) for more details.",
```
serdardalgic, line 25: You would probably refer to [Managing Replicas EDB Doc](https://www.enterprisedb.com/docs/biganimal/latest/using_cluster/managing_replicas/) in the description.

serdardalgic (changes requested):
Thanks for your work @edbamolvyavahare, It looks good in general. I wrote some comments. 

One general idea about the `cluster_type`: I think we shouldn't make it an optional field, but rather a computed field. The `biganimal_cluster` resource will only create `cluster` cluster_type, and the `biganimal_faraway_replica` resource would only create `faraway_replica` types. We shouldn't allow creation of faraway replicas through biganimal_cluster (if that's the case, we don't need a new resource called `biganimal_faraway_replica` 😅). That is basically possible, but I think there is a certain value in this level of separation for now, I mean having two distinct resources: `biganimal_cluster` and `biganimal_faraway_replica`.

pkg/models/fa_replica.go
```diff
@@ -0,0 +1,150 @@
+package models
+
+import (
+	"github.com/EnterpriseDB/terraform-provider-biganimal/pkg/utils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+//const (
```
YanniHu1996, line 8: remove useless code

examples/resources/biganimal_fa_replica/aws/resource.tf
edbamolvyavahare: Sounds good, fixed.

pkg/provider/resource_cluster.go
edbamolvyavahare: missed this, fixed it.

edbamolvyavahare:
> Thanks for your work @edbamolvyavahare, It looks good in general. I wrote some comments.
> 
> One general idea about the `cluster_type`: I think we shouldn't make it an optional field, but rather a computed field. The `biganimal_cluster` resource will only create `cluster` cluster_type, and the `biganimal_faraway_replica` resource would only create `faraway_replica` types. We shouldn't allow creation of faraway replicas through biganimal_cluster (if that's the case, we don't need a new resource called `biganimal_faraway_replica` 😅). That is basically possible, but I think there is a certain value in this level of separation for now, I mean having two distinct resources: `biganimal_cluster` and `biganimal_faraway_replica`.

Agreed, Fixed

pkg/models/fa_replica.go
edbamolvyavahare: done

pkg/provider/resource_fareplica.go
edbamolvyavahare: Thanks, done.

examples/resources/biganimal_faraway_replica/azure/resource.tf
```diff
@@ -0,0 +1,88 @@
...
+  //  pg_config {
+  //    name  = "application_name"
+  //    value = "created through terraform"
+  //  }
+  //
+  //  pg_config {
+  //    name  = "array_nulls"
+  //    value = "off"
+  //  }
+
```
edbamolvyavahare, line 58:
```suggestion

```

serdardalgic (changes requested):
So, 
* I tested creating a FAR from a cluster. The code creates a FAR from an existing cluster. 
* However, the code doesn't recognize whether the FAR cluster is promoted or not. If the FAR cluster is promoted, the following error occurs when I try to `terraform apply`:
```
biganimal_faraway_replica.faraway_replica: Modifying... [id=p-j6j5pjx391]
╷
│ Error: status: 400 - Bad Request
│
│   with biganimal_faraway_replica.faraway_replica,
│   on resource.tf line 29, in resource "biganimal_faraway_replica" "faraway_replica":
│   29: resource "biganimal_faraway_replica" "faraway_replica" {
│
│ At least one of 'replicas' and 'clusterArchitecture' must be specified.
```
Because it's no longer a FAR cluster. 
Before making an update call to our BA API, we need to detect whether this cluster is a faraway replica or not. 

Another thing I checked was giving a wrong source_id:
* if there is no cluster with the given source_cluster_id, API returns `404 - Not Found, no such cluster`
* If the source ID is a FAR cluster, API returns `400 - Bad Request, Source cluster can not be replica`

pkg/provider/resource_cluster.go
serdardalgic: Btw, I tested this scenario and we don't need the state migration, because we're setting the cluster_type during the read function.  

I'm resolving this conversation.

pkg/models/fa_replica.go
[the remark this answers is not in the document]
edbamolvyavahare: Done & verified.

serdardalgic (approved):

PR #148, opened 2023-03-01
style: add vendor to gitignore

serdardalgic:
We're not using any vendor directory. In order to prevent adding it mistakenly, let's add vendor to gitignore. Please review.


YanniHu1996 (approved):
LGTM

PR #149, opened 2023-03-02
feat: add support for faraway replica data source

edbamolvyavahare:


pkg/provider/data_source_fareplica.go
```diff
@@ -0,0 +1,300 @@
...
+type FAReplicaData struct{}
+
+func NewFAReplicaData() *FAReplicaData {
+	return &FAReplicaData{}
+}
+
+//
+func (c *FAReplicaData) Schema() *schema.Resource {
+	return &schema.Resource{
+		Description: "The cluster data source describes a BigAnimal cluster. The data source requires your cluster name.",
```
YanniHu1996, line 24: Update `description`

YanniHu1996 (commented):
How to ensure that the cluster from this data_source is `faraway_replica` cluster? I think `client.ReadByName` can't ensure that
I think a better way is calling `/projects/{projectId}/clusters/{clusterId}/faraway-replicas` to get a cluster's list of a source cluster.
So the input should be `source_cluster_id` not `cluster_name`

pkg/provider/data_source_fareplica.go
edbamolvyavahare: Done

pkg/provider/data_source_fareplica.go
```diff
@@ -0,0 +1,300 @@
...
+		},
+	}
+}
+
+//
+func (c *FAReplicaData) Read(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	client := api.BuildAPI(meta).ClusterClient()
+
+	clusterName, ok := d.Get("cluster_name").(string)
```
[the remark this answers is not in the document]
edbamolvyavahare: Done

edbamolvyavahare:
> How to ensure that the cluster from this data_source is a `faraway_replica` cluster? I think `client.ReadByName` can't ensure that I think a better way is calling `/projects/{projectId}/clusters/{clusterId}/faraway-replicas` to get a cluster's list of a source cluster. So the input should be `source_cluster_id` instead of `cluster_name`

Agree, taking this into consideration, we've planned to provide two options to users to fetch faraway replica cluster details, 
1. There will be in total 2 data-sources: `biganimal_cluster_faraway_replicas`  and `biganimal_faraway_replica`, `biganimal_cluster_faraway_replicas`, this returns a list of faraway replicas. Input is source cluster ID of cluster_type=cluster. i.e Plural response

2. `biganimal_faraway_replica`. This fetches info of a single FA-Replica. fetch by FAR cluster-ID. SINGULAR. It will only return the value when cluster_type=faraway_replica

pkg/provider/validators.go
```diff
@@ -51,3 +51,17 @@ func validateUUID(v interface{}, _ cty.Path) diag.Diagnostics {
 	}
 	return nil
 }
+
+func validateClusterName(v interface{}, _ cty.Path) diag.Diagnostics {
```
serdardalgic, line 55: I think we don't need to do this validation. 

API will already return a 400 - Bad Request, when there is no name provided for the biganimal_cluster data-source. 
Screenshot 2023-03-24 at 08 57 41
WDYT, @edbamolvyavahare ?

serdardalgic (commented):
So, I tested the implementation, and so far, it looks good. 
* I created a single_node cluster `serdarFATest` with the TF provider
* Then I created a faraway replica in the BigAnimal portal. 
* I tried to read the `serdarFATest` with `biganimal_cluster` data-source, it worked fine. 
* I tried to read the faraway replica with `biganimal_cluster` data-source, and it gave an error as expected:
Screenshot 2023-03-24 at 08 30 23
* I tried to read the cluster with `biganimal_faraway_replica` data-source, and it gave and error as expected.
Screenshot 2023-03-24 at 08 34 24
* Finally, I was able to read the faraway replica with the `biganimal_faraway_replica` data-source, worked perfect. 
* Then I promoted the faraway replica:
  * serdarFATest `biganimal_cluster` resource has no faraway replica any more, it's shown as expected in the data-source `terraform plan` output.
  * The `biganimal_faraway_cluster` resource is no longer a faraway replica any more, and it was treated as a regular cluster. 

There is only one minor decision part about the validation for the cluster name, let's decide it and merge this PR then @edbamolvyavahare , I think it's good to go.

pkg/provider/validators.go
YanniHu1996: If the "name" argument is validated during the input phase, it would be beneficial to promptly alert users of any errors.

YanniHu1996 (approved):
Thanks effort @edbamolvyavahare  ! LGTM

PR #144, opened 2023-02-27
feat: Support CSP connection

YanniHu1996:
This PR implemented the following resource
- resource `aws_connection` 
- resource `azure_connection`
- data_resource `aws_connection` 

Due to functional and API limitations, the `delete` method in `aws_connection` has not been implemented, `update`,`read`,`delete` method int `azure_connection` has not been implemented.

PR #154, opened 2023-03-15
Fix missing checks

serdardalgic:
Some checks didn't run while we were reviewing the #144, such as spellchecker. 

That's why I'm creating this PR.


edbamolvyavahare (approved):
LGTM

PR #160, opened 2023-03-17
fix: make spellchecker run on a branch in fork repo

YanniHu1996:


serdardalgic (dismissed):
LGTM. Thank you @YanniHu1996 

Do you know why [the pre-commit workflow](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/.github/workflows/pre-commit.yml) is not triggered for this PR? Should we also trigger it for the branches that come from a fork repo? What do you think? (If yes, we can tackle it in another PR, I think.) 

So far, this PR looks good to go, IMHO.

serdardalgic (approved):
As can be seen from this PR, the checks are working.

PR #170, opened 2023-03-23
ci: Update golangci-lint and golang versions

serdardalgic:
Housekeeping, let's update the Golang and Golangci-lint versions


YanniHu1996 (approved):
Thanks!

PR #177, opened 2023-03-28
test: Add Acceptance Tests

serdardalgic:
This PR is still a Draft:

Here is an initial effort for `biganimal_cluster` resource Acceptance test. 

You can run the test with the following command
```
TF_ACC=1 go test -timeout 600s -run ^TestAccResourceCluster$ github.com/EnterpriseDB/terraform-provider-biganimal/pkg/provider
```
or, with this one for more verbosity:
```
TF_ACC_LOG=DEBUG TF_ACC=1 go test -v -timeout 600s -run ^TestAccResourceCluster$ github.com/EnterpriseDB/terraform-provider-biganimal/pkg/provider
```

The Plan Change after the initial `terraform apply` is suppressed, because of the pg_config and allowedIpRanges coming with different order. 

Additional Links:
https://developer.hashicorp.com/terraform/plugin/testing/testing-patterns


pkg/provider/resource_region_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"github.com/EnterpriseDB/terraform-provider-biganimal/pkg/api"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccResourceRegion_basic(t *testing.T) {
+	var (
+		acc_env_vars_checklist = []string{
+			"BA_TF_ACC_VAR_region_project_id",
```
YanniHu1996, line 15: a minor comment, we can remove the common prefix `BA_TF_VAR` for reducing the variable length

serdardalgic: I wanted to have the ACC_VAR inside the env var name, to explicitly show that this var will be used for the Acceptance tests. My concern if we remove the ACC from the env var's name, and use BA_TF_VAR , people might confuse it with the TF_VARs that they can use while running terraform commands, like:
```
TF_VAR_cluster_name=<cluster_name> terraform plan
```
I'd vote for having it BA_TF_ACC_VAR to be explicit about the env var name.

YanniHu1996 (approved):
LGTM

YanniHu1996 (approved):

PR #185, opened 2023-04-03
refactor: rewrite `resource_projects` and `datasource_projects` with terraform-plugin-framework

YanniHu1996:


pkg/provider/provider.go
```diff
@@ -86,13 +78,122 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
 		ba_bearer_token := schema.Get("ba_bearer_token").(string)
 		ba_api_uri := schema.Get("ba_api_uri").(string)
 		// set our meta to be a new api.API
-		// this can be turned into concrete clients
+		// this can be turned into concrete cba_api_uri := schema.Get("ba_api_uri").(string)lients
```
serdardalgic, line 81: This line is probably a copy paste error, right?

pkg/provider/provider.go
```diff
@@ -86,13 +78,122 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
...
+		diags := diag.Diagnostics{}
+
+		if ba_bearer_token == "" {
+			ba_bearer_token = os.Getenv("BA_BEARER_TOKEN")
+		}
+
+		if ba_bearer_token == "" {
+			diags = append(diags, diagv2.Diagnostic{
+				Severity: diagv2.Error,
+				Summary:  "Unable to find ba_nearer_token",
```
serdardalgic, line 97:
```suggestion
				Summary:  "Unable to find ba_bearer_token",
```

pkg/provider/resource_proejct.go
```diff
@@ -0,0 +1,231 @@
+package provider
```
serdardalgic, line 1: There is a typo in the filename, must be `resource_project.go`, it is now `resource_proejct.go`

pkg/provider/resource_proejct_test.go
```diff
@@ -0,0 +1,42 @@
+package provider
```
serdardalgic, line 1: Also, the filename must be corrected here.

pkg/provider/provider.go
YanniHu1996: sure, fixed

pkg/models/project.go
```diff
@@ -1,16 +1,16 @@
...
 type Project struct {
-	ProjectId      string          `json:"projectId,omitempty" mapstructure:"project_id"`
-	ProjectName    string          `json:"projectName,omitempty" mapstructure:"name"`
-	UserCount      int             `json:"userCount,omitempty" mapstructure:"user_count"`
-	ClusterCount   int             `json:"clusterCount,omitempty" mapstructure:"cluster_count"`
-	CloudProviders []CloudProvider `json:"cloudProviders" mapstructure:"cloud_providers"`
+	ProjectId      string           `json:"projectId,omitempty" tfsdk:"project_id"`
+	ProjectName    string           `json:"projectName,omitempty" tfsdk:"project_name"`
+	UserCount      int              `json:"userCount,omitempty" tfsdk:"user_count"`
+	ClusterCount   int              `json:"clusterCount,omitempty" tfsdk:"cluster_count"`
```
serdardalgic, line 12: Here, the linter is confused, I guess. Also, the precommit hook is complaining about this indentation. 
```suggestion
	ProjectId      string          `json:"projectId,omitempty" tfsdk:"project_id"`
	ProjectName    string          `json:"projectName,omitempty" tfsdk:"project_name"`
	UserCount      int             `json:"userCount,omitempty" tfsdk:"user_count"`
	ClusterCount   int             `json:"clusterCount,omitempty" tfsdk:"cluster_count"`
```

pkg/provider/provider.go
```diff
@@ -3,23 +3,26 @@ package provider
...
-
 	"github.com/EnterpriseDB/terraform-provider-biganimal/pkg/api"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	schema2 "github.com/hashicorp/terraform-plugin-framework/provider/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	diagv2 "github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"os"
```
serdardalgic, line 14: `"os"` should go to the beginning of the imports. I guess this is the difference between goimports and gofmt.

pkg/provider/provider.go
```diff
@@ -93,6 +85,115 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
...
+
+	if token == "" {
+		resp.Diagnostics.AddError(
+			"Unable to find ba_nearer_token",
+			"ba_nearer_token cannot be an empty string",
+		)
+		return
+	}
+
+	var host = "https://portal.biganimal.com/api/v3"
```
serdardalgic, line 159: Should we write this as a constant in this package and use it everywhere we need it?

pkg/provider/resource_project.go
```diff
@@ -80,73 +96,136 @@ func (p *ProjectResource) Schema() *schema.Resource {
...
 
+	projectId, err := p.client.ProjectClient().Create(ctx, config.ProjectName.ValueString())
 	if err != nil {
-		return fromBigAnimalErr(err)
+		resp.Diagnostics.AddError("Error creating project", "Could not create project, unexpected error: "+err.Error())
+		return
 	}
 
-	d.SetId(projectId)
+	project, err := p.client.ProjectClient().Read(ctx, projectId)
```
serdardalgic, line 137: As a general idea, can we move this `p.client.ProjectClient()` to the Configure function, where we set the provider configured client? 🤔 So that, we can call the client functions as:
```
p.client.Create
```
It looks more readable to me, rather than `p.client.ProjectClient().Create`. 
We shouldn't initialize a ProjectClient every time we call a method, IMHO.

pkg/provider/provider.go
```diff
@@ -93,6 +85,115 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
...
+		}
+	}
+}
+
+func (b bigAnimalProvider) Metadata(_ context.Context, _ provider.MetadataRequest, resp *provider.MetadataResponse) {
+	resp.TypeName = "biganimal"
+	resp.Version = b.version
+}
+
+func (b bigAnimalProvider) Configure(ctx context.Context, req provider.ConfigureRequest, resp *provider.ConfigureResponse) {
```
serdardalgic, line 136: Can you write unit tests for this Configure function, to ensure that we cover all the cases. It would be helpful to check cases like:
1. Those two fields are set in the provider block
2. One of them are set in the provider block
3. They are not set in the provider block, but present as environment variables
4. no env vars, not defined within a provider block, we expect diag error.

serdardalgic (commented):
Overall, looks good. I think we're on a good path here. 

I left some comments. Also, keep in mind, the data source is plural `biganimal_projects` but the resource is singular `biganimal_project`. 

The data source returns a List of projects. The resource tackles with only one object.

pkg/provider/resource_project.go
YanniHu1996: that's great, follow your thought

main.go
```diff
@@ -29,16 +36,42 @@ var (
...
+			return schema.NewGRPCProviderServer(provider.New(version)())
+		},
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
 
-		ProviderAddr: "registry.terraform.io/EnterpriseDB/biganimal",
+	providers := []func() tfprotov6.ProviderServer{
+		providerserver.NewProtocol6(provider.NewProvider(version)()),
```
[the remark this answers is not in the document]
YanniHu1996: thanks for your suggestion, I change new functions name `NewFrameworkProvider` and `NewSDKProvider` respectively

pkg/provider/provider.go
YanniHu1996: thanks! Extracted a constant

pkg/provider/resource_project.go
```diff
@@ -80,73 +98,140 @@ func (p *ProjectResource) Schema() *schema.Resource {
...
+	}
 
-	if err := p.read(ctx, d, meta); err != nil {
-		return diag.FromErr(err)
+	config.ID = types.StringValue(project.ProjectId)
+	config.ProjectID = types.StringValue(project.ProjectId)
+	config.ProjectName = types.StringValue(project.ProjectName)
+	config.UserCount = types.Int64Value(int64(project.UserCount))
+	config.ClusterCount = types.Int64Value(int64(project.ClusterCount))
+	config.CloudProviders = []cloudProvider{}
```
serdardalgic, line 150: When we create a project, the API is returning a `set` of cloud_provider values. We shouldn't set the `config.CloudProviders` to an empty list here, but we should read the returned values from the API and set them here.

pkg/provider/resource_project.go
```diff
@@ -80,73 +98,140 @@ func (p *ProjectResource) Schema() *schema.Resource {
...
+		resp.Diagnostics.AddError("Error reading project", "Could not read project, unexpected error: "+err.Error())
+		return
+	}
+
+	state.ProjectName = types.StringValue(project.ProjectName)
+	state.UserCount = types.Int64Value(int64(project.UserCount))
+	state.ClusterCount = types.Int64Value(int64(project.ClusterCount))
+	if cps := project.CloudProviders; cps != nil {
+		for _, provider := range cps {
+			state.CloudProviders = append(state.CloudProviders, cloudProvider{
```
serdardalgic, line 179: project.CloudProviders is of type `Set` (https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types/basetypes#SetType), not `List` because they're an unordered collection of unique objects.
so, plainly appending the read values will not work. 

The Read function should read the project.CloudProviders values and if they do not exist in `state.CloudProviders`, they should be added to the set.

serdardalgic (changes requested):
While dealing with TypeSet type of CloudProviders set, we need to update the login during read and create. Here I left some comments.

pkg/provider/resource_project.go
```diff
@@ -2,76 +2,93 @@ package provider
...
-				Computed:    true,
+			"user_count": schema.Int64Attribute{
+				MarkdownDescription: "User Count of the project.",
+				Computed:            true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+			"cluster_count": schema.Int64Attribute{
+				MarkdownDescription: "User Count of the project.",
```
serdardalgic, line 59: Should be cluster count.

pkg/provider/resource_project.go
```diff
@@ -80,73 +97,145 @@ func (p *ProjectResource) Schema() *schema.Resource {
...
+		resp.Diagnostics.AddError("Error reading project", "Could not read project, unexpected error: "+err.Error())
+		return
+	}
+
+	state.ProjectID = &project.ProjectId
+	state.ProjectName = &project.ProjectName
+	state.UserCount = &project.UserCount
+	state.ClusterCount = &project.ClusterCount
+	state.CloudProviders = nil
+	for _, provider := range project.CloudProviders {
```
serdardalgic, line 183: This code block will add the cloud providers even if they are written to the state. We need to check if they already exist in the cloud providers section

serdardalgic (changes requested):
Thanks for the changes. Can you please take a look at the read section again? It looks like in every read, we are trying to add the cloud provider values to that section. That would break the provider with duplicate values error.

Issue #88, opened 2022-11-23
Implement Import for biganimal_cluster resource


serdardalgic:
Currently, we don't have import support for `biganimal_cluster`. 

Import for that resource should be implemented, tested and documented.

serdardalgic:
I tackled with this issue a bit today. 
I've added the following part to the schema of the `biganimal_cluster` resource:
```
    Importer: &schema.ResourceImporter{
       StateContext: schema.ImportStatePassthroughContext,
    },
```
and tried running the following command:
```
$> terraform import biganimal_cluster.single_node_cluster <cluster_id>
```
on [examples/resources/biganimal_cluster/single_node/aws](https://github.com/EnterpriseDB/terraform-provider-biganimal/tree/main/examples/resources/biganimal_cluster/single_node/aws) directory. 

What ImportStatePassthroughContext does is passing the resource id to the read function of that resource, and then read the resource. 

However, we also `project_id` to read the cluster resource, that's why it fails. 

I tried running the import command like this:
```
terraform import -var project_id=<project_id> -var cluster_name=<cluster_name> biganimal_cluster.single_node_cluster <cluster_id>
```
but it doesn't pass the project_id to the read function. I cannot get it with 
```
projectId := d.Get("project_id").(string)
```
. 

If anyone can try it and find a way to get the project_id, then we can implement the importer function. 
The way import functions is similar in every resource, so, once we get it done for `biganimal_cluster` resource, we can use it for `biganimal_project` and `biganimal_region` resources as well. 

Or, an alternative way to implement import would be changing the ID of the resources. We wouldn't use `cluster_id` as the resource ID, but `/projects/{projectId}/clusters/{clusterId}` so that users would import `biganimal_cluster` resources like:
```
terraform import biganimal_cluster.this /projects/{projectId}/clusters/{clusterId}
```
But that means, we need to change the way we store the terraform state. 

It makes sense to me, but I'm eager to hear if you have any suggestions.

PR #187, opened 2023-04-04
docs: Mostly whitespace changes in the docs

serdardalgic:
When we're using templates, it's good to keep our code as close to [the upstream code](https://github.com/hashicorp/terraform-plugin-docs/blob/e8f8eb1f6dbb047551ceedd48e7810c6c6661d03/internal/provider/template.go#L217) as possible. 

The template doesn't work, when I use the 
```
{{ printf "{{tffile %q}}" .ExampleFile }}
```
I don't know why. That's why I kept on using 
```
{{ tffile .ExampleFile }}
```
Same goes for 
```
{{ codefile "shell" .ImportFile }}
```
(anyhow, we don't have the import functionality right now, see #88 for details)


YanniHu1996 (approved):
LGTM

PR #191, opened 2023-04-05
docs: Add development related docs

### LINKED ISSUE #190: Document easier token-fetch by using BigAnimal CLI
**Is your feature request related to a problem? Please describe.**
We're describing how to get the Auth Token with the get-token.sh script. It's a cumbersome process, and developers themselves don't even use that script. 

**Describe the solution you'd like**
BigAnimal CLI is quite handy in managing secrets. We can describe how to use BigAnimal CLI to manage the auth token for different environments.  

**Describe alternatives you've considered**
Until #41 is implemented, documenting BigAnimal CLI usage to manage the tokens would be helpful.

**Additional context**
https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#authenticate-as-a-valid-user

serdardalgic:
* Added a new ./DEVELOPMENT.md file and moved the testing and development related information there. 
* Also added how to use the BigAnimal CLI for easier token management, Fixes #190. 

Please review.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,143 @@
+# Development
+
+Please make sure to read the [Contributing guideline](./CONTRIBUTING.md) first.
+
+## How to manage the BigAnimal Bearer Token?
+
+As described in the [Readme](./README.md#getting-an-api-token) file, you can use the `get-token.sh` script to fetch a Bearer Token.
+
+Another common practice is using the [Biganimal CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/).
+
+### Using BA CLI to help initializing Provider credentials
+
+1. [Install the BA CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli) and [jq - Command Line JSON Processor ](https://stedolan.github.io/jq/).
+1. [Authenticate as a valid user and create a credential](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli). This command will direct you to your browser.
+```shell
+biganimal create-credential \
+  --name "ba-user1" \
+  --address "portal.biganimal.com" \
+  --port "443"
+```
+3. Add the following bash functions to your shellrc file (For example: `.bashrc` if you're using bash, `.zshrc` if you're using ZSH) and start a new shell.
+```bash
+ba_api_get_call () {
+	endpoint=$1
+	curl -s --request GET --header "content-type: application/json" --header "authorization: Bearer $BA_BEARER_TOKEN" --url "$BA_API_URI$endpoint"
+}
+
+ba_get_proj_id () {
+	echo $(ba_api_get_call "/user-info" | jq -r ".data.organizationId" | cut -d"_" -f2)
+}
+
+export_BA_token_from_CLI () {
+	cred_name="${1:-ba-user1}" ## Replace "ba-user1" with your credential name, if you're using something different
+	if ! biganimal show-clusters -c $cred_name > /dev/null
+	then
+		echo "!!! Running the reset-credential command now !!!"
+		biganimal reset-credential $cred_name
+	fi
+	biganimal show-clusters -c $cred_name >&/dev/null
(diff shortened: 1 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Using BA CLI to help initializing Provider credentials
+
+1. [Install the BA CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli) and [jq - Command Line JSON Processor ](https://stedolan.github.io/jq/).
+1. [Authenticate as a valid user and create a credential](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli). This command will direct you to your browser.
+```shell
+biganimal create-credential \
+  --name "ba-user1" \
+  --address "portal.biganimal.com" \
+  --port "443"
```
YanniHu1996, line 19: The default values for address and port are "portal.biganimal.com" and "443", respectively. Hence, it's not mandatory to include them in the command "biganimal create-credential --name "ba-user"".

DEVELOPMENT.md
```diff
@@ -0,0 +1,143 @@
...
+```
+3. Add the following bash functions to your shellrc file (For example: `.bashrc` if you're using bash, `.zshrc` if you're using ZSH) and start a new shell.
+```bash
+ba_api_get_call () {
+	endpoint=$1
+	curl -s --request GET --header "content-type: application/json" --header "authorization: Bearer $BA_BEARER_TOKEN" --url "$BA_API_URI$endpoint"
+}
+
+ba_get_proj_id () {
```
YanniHu1996, line 28: The organization ID is set as the default project, therefore I believe the most fitting function name would be `ba_get_default_proj_id`.

DEVELOPMENT.md
```diff
@@ -0,0 +1,143 @@
...
+	echo $(ba_api_get_call "/user-info" | jq -r ".data.organizationId" | cut -d"_" -f2)
+}
+
+export_BA_token_from_CLI () {
```
YanniHu1996, line 32: suggestion function name "export_BA_env",

DEVELOPMENT.md
serdardalgic: This is the default example we give in [the EDB Docs](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#authenticate-as-a-valid-user) as well, that's why I wanted to keep it aligned with our official docs. 

But true, they are optional. I'm in favor of keeping them, if you don't have a strong objection.

DEVELOPMENT.md
```diff
@@ -0,0 +1,143 @@
...
+## How to manage the BigAnimal Bearer Token?
+
+As described in the [Readme](./README.md#getting-an-api-token) file, you can use the `get-token.sh` script to fetch a Bearer Token.
+
+Another common practice is using the [Biganimal CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/).
+
+### Using BA CLI to help initializing Provider credentials
+
+1. [Install the BA CLI](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli) and [jq - Command Line JSON Processor ](https://stedolan.github.io/jq/).
+1. [Authenticate as a valid user and create a credential](https://www.enterprisedb.com/docs/biganimal/latest/reference/cli/#installing-the-cli). This command will direct you to your browser.
```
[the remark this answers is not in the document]
serdardalgic: Github Markdown automatically increases the number, until there is a code sample. You can check the rich display. 
So, this `1.` is interpreted as the next number, which is `2.` here :)

DEVELOPMENT.md
serdardalgic: That's a good idea, thanks 👍

DEVELOPMENT.md
serdardalgic: yeah, because it's not only the token. How about `export_BA_env_vars`?

DEVELOPMENT.md
YanniHu1996: Sure, I agree with you

YanniHu1996 (approved):
LGTM

PR #206, opened 2023-04-20
feat: throughput storage field could be configurable

### LINKED ISSUE #205: the `throughput`  field of the `storage` section could be configurable
Since the last release, the throughput field could be provided as input param by the user

Screenshot 2023-04-20 at 09 49 56

https://status.biganimal.com/maintenance-window/3I1V4U2

mbiagetti:
Define `throughput` field as optional

will close #205

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -336,9 +336,6 @@ Required:
 Optional:
 
 - `iops` (String) IOPS for the selected volume. It can be set to different values depending on your volume type and properties.
-
-Read-Only:
-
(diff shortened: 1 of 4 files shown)


serdardalgic:
One thing we forgot to implement here, @mbiagetti 
When the throughput value is not set during creation time, it will be computed. But the next time we run `terraform plan/apply`, it will show a diff. 

In order to mitigate this, we need to write [a DiffSuppressFunc](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/pkg/provider/resource_cluster.go#L236-L242) like the iops field. 

I will raise a PR for that.

PR #128, opened 2023-02-01
fix: Make allowedIpRanges and pgConfig TypeSet

serdardalgic:
These changes are taken out of #123, and they will fix #98. 

As explained [in this comment](https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/123#issuecomment-1412116738), they're breaking changes that demand a major bump. 

That's why I would like to keep this PR here and merge it before a major bump (unless we decide to break the API while we still don't have many active users that would be affected from this change)

PR #216, opened 2023-04-27
fix: reflect TypeSet change to FAR as well

serdardalgic:
Those changes were forgotten when we merged #128. 
Faraway replicas can not be created with 0.4.1 as of now, so, this bug fix is important 😅 

Please review.


edbamolvyavahare (approved):
LGTM

Issue #131, opened 2023-02-08
`biganimal_region` resources should be renamed as `biganimal_regions` 


serdardalgic:
Although it's returning a List, the name of the data-source is singular. 

That's against [our Naming Conventions](https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/CONTRIBUTING.md#naming). 

Changing this would be a breaking change. Please refer to [Provider Data Source or Resource Rename Documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-rename) for further details. Quoting there:

> When renaming a resource from one name to another, it is important to keep backwards compatibility with both existing Terraform configurations and the Terraform state while operators migrate. To accomplish this, there will be some duplicated logic to support both resources until the next MAJOR release. Once both resources are appropriately handled, the process for deprecating and removing the old resource is the same as noted in the [Provider Data Source or Resource Removal section](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal).

PR #230, opened 2023-05-15
Refactor/migrate region

YanniHu1996:


pkg/provider/provider.go
```diff
@@ -181,11 +177,13 @@ func (b bigAnimalProvider) DataSources(ctx context.Context) []func() datasource.
 	return []func() datasource.DataSource{
 		NewProjectsDataSource,
 		NewPgdDataSource,
+		NewRegionDataSource,
```
serdardalgic, line 180: Like the `NewProjectsDataSource` we need to rename this resource as `NewRegionsDataSource` as the data source is returning a list of regions, instead of a single region output. See #131

serdardalgic (commented):
I merged the main into the working branch, and made some changes in the current code. 
Right now, the code gets compiled. But I haven't tested it thoroughly.

pkg/provider/provider.go
YanniHu1996: ok, I will follow this design

serdardalgic: That's actually not enough :) 
We need to rename the data source from `biganimal_region` to `biganimal_regions`, update the examples and the documentation, and also mark the `biganimal_region` data source as deprecated. 

I wonder if we should do all those in this PR, or create a new one. I'm fine with both options.

pkg/provider/data_source_regions.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (r *regionDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_region"
+}
+
+func (r *regionDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	r.regionsDataSource.Schema(ctx, req, resp)
+	resp.Schema.DeprecationMessage = "The datasource' 'region' is deprecated and will be removed in the next major version. Please use 'regions' instead."
+}
```
serdardalgic, line 153: How about moving this part to the `pkg/provider/data_source_region.go` instead of removing that file?

pkg/provider/data_source_regions_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccDataSourceRegions_basic(t *testing.T) {
+	var (
+		acc_env_vars_checklist = []string{
+			"BA_TF_ACC_VAR_datasource_regions_project_id",
+			"BA_TF_ACC_VAR_datasource_regions_provider",
+			"BA_TF_ACC_VAR_datasource_regions_region_id",
+			"BA_TF_ACC_VAR_datasource_regions_status",
```
serdardalgic, line 17: Please add those newly added ACC Test env vars to the `.env.example` file.

serdardalgic: For the environment variable names, I'd suggest going with e.g. `BA_TF_ACC_VAR_regions_project_id`. We can skip the `_datasource_`. 

This is how we did for the PGD Data source test. 
If you would like to keep the `_datasource_` in the acceptance test env var names, let's update other data source env var names as well (preferably in a separate PR)

pkg/provider/data_source_regions_test.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		projectId  = os.Getenv("BA_TF_ACC_VAR_datasource_regions_project_id")
+		provider   = os.Getenv("BA_TF_ACC_VAR_datasource_regions_provider")
+		regionId   = os.Getenv("BA_TF_ACC_VAR_datasource_regions_region_id")
+		regionName = os.Getenv("BA_TF_ACC_VAR_datasource_regions_region_name")
+		status     = os.Getenv("BA_TF_ACC_VAR_datasource_regions_status")
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
```
serdardalgic, line 26: A suggestion would be:
* Let's use a `biganimal_region` resource and read this region data with the data source, and then compare their values with `resource.TestCheckResourceAttrPair`. Similar to how it's described in [the Terraform AWS Provider's Data source Acceptance test documentation](https://hashicorp.github.io/terraform-provider-aws/running-and-writing-acceptance-tests/#data-source-acceptance-testing). What do you think, @YanniHu1996?

If you think it would take a long time, please create a GitHub Issue to improve the Data source acceptance tests. We can tackle it in the future.

pkg/provider/resource_region.go
```diff
@@ -4,185 +4,243 @@ import (
...
+				},
 			},
-			"continent": {
-				Description: "Continent that region belongs to. For example, \"Asia\", \"Australia\", or \"Europe\".",
-				Type:        schema.TypeString,
-				Computed:    true,
+			"status": schema.StringAttribute{
+				MarkdownDescription: "Region status of the region. For example, \"ACTIVE\", \"INACTIVE\", or \"SUSPENDED\".",
+				Optional:            true,
+				/*
```
serdardalgic, line 75: What will we do with this default part? @YanniHu1996

pkg/provider/resource_region.go
```diff
@@ -4,185 +4,243 @@ import (
...
-	return diag.Diagnostics{}
 }
 
-func (r *RegionResource) retryFunc(ctx context.Context, d *schema.ResourceData, meta any, cloudProvider, regionId, desiredState string) retry.RetryFunc {
-	client := api.BuildAPI(meta).RegionClient()
+func (r *regionResource) retryFunc(ctx context.Context, region Region) retry.RetryFunc {
 	return func() *retry.RetryError {
-		projectId := d.Get("project_id").(string)
-		region, err := client.Read(ctx, projectId, cloudProvider, regionId)
+		curr, err := r.client.RegionClient().Read(ctx, *region.ProjectID, *region.CloudProvider, *region.RegionID)
```
serdardalgic, line 217:
```suggestion
		curr, err := r.client.Read(ctx, *region.ProjectID, *region.CloudProvider, *region.RegionID)
```

pkg/provider/resource_region.go
```diff
@@ -4,185 +4,243 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/path"
```
serdardalgic, line 7: If I'm not mistaken, this should be added to the down (in the following lines of the import statement)

templates/data-sources/regions.md.tmpl
serdardalgic, line 1: We probably don't need `regions.md.tmpl`. 
As I mentioned above, we can keep `region.md.tmpl` with a deprecation notice at the top of the template.

pkg/provider/validators.go
```diff
@@ -28,6 +31,13 @@ func validateProjectId(v interface{}, path cty.Path) diag.Diagnostics {
 	return diags
 }
 
+func ProjectIdValidator() validator.String {
```
serdardalgic, line 34: It would be nice to check if we can add this validator to the other parts of the code. It would be great to have a GitHub Issue for it, and then we can add this validator to other parts of the code in a separate PR. 

This function would stay here in this PR. Up to you.

serdardalgic (changes requested):
Thanks for your work, @YanniHu1996. The PR is going in a good direction, IMHO. 
I've requested some changes. 
After you fix them, please check the following:
* Please make sure to run a `make docs` for the latest documentation. 
* Please ensure that the examples are correct. In the examples, we can add `bah:aws` to the `cloud_provider` variables, for example the `biganimal_regions` data source. 
* please test the `biganimal_region` and `biganimal_regions` data sources, working as expected. (`biganimal_region` giving a deprecation notice). 
* please test the `biganimal_region` resource, preferably with the acceptance tests. If you need to introduce new tests for the `biganimal_region` resource other than the Basic test, we can include it within this PR. 

Thanks a lot!

pkg/provider/data_source_regions_test.go
YanniHu1996: I will remove `datasource_` part

pkg/provider/resource_region.go
YanniHu1996: sorry, this is wrong comment, I will recover

pkg/provider/resource_region.go
YanniHu1996: right, rerun `goimports`

pkg/provider/data_source_regions.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+			"region_id": schema.StringAttribute{
+				Description: "Unique region ID. For example, \"germanywestcentral\" in the Azure cloud provider, \"eu-west-1\" in the AWS cloud provider.",
+				Optional:    true,
+			},
+		},
+	}
+}
+
+type regionDatasource struct {
```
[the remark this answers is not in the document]
YanniHu1996: I updated to `regionsDataSourceModel`

pkg/provider/validators.go
YanniHu1996: Thanks! I prefer to keep this function in this PR

templates/data-sources/regions.md.tmpl
YanniHu1996: I think we don't need `region.md.tmpl`. I will run `tfplugindocs` generate --ignore-deprecated to avoid generating documentation for the data_source regions

pkg/provider/utils.go
```diff
@@ -3,7 +3,10 @@ package provider
 import (
 	"errors"
 
+	"fmt"
+
 	"github.com/EnterpriseDB/terraform-provider-biganimal/pkg/api"
+	diag2 "github.com/hashicorp/terraform-plugin-framework/diag"
```
serdardalgic, line 9:
```suggestion
	frameworkdiag "github.com/hashicorp/terraform-plugin-framework/diag"
```
Instead of diag2, let's be more specific on where this diag comes from, as we did in provider.go.
Please update the fromErr function accordingly.

pkg/provider/utils.go
```diff
@@ -3,7 +3,10 @@ package provider
...
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
serdardalgic, line 11:
```suggestion
	sdkdiag "github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
As we're introducing frameworkdiag, let's name this one as sdkdiag, and update the fromBigAnimalErr function as well. Thank you 🙏

pkg/provider/data_source_regions.go
```diff
@@ -0,0 +1,134 @@
...
+		if err != nil {
+			resp.Diagnostics.Append(fromErr(err, "Error reading region by id: %v", cfg.RegionId)...)
+			return
+		}
+		regions = append(regions, region)
+
+	} else {
+		respRegions, err := r.client.List(ctx, *cfg.ProjectId, *cfg.CloudProvider, cfg.Query.ValueString())
+		if err != nil {
+			return
```
serdardalgic, line 125: There is something wrong in this part of the code. 
We call the `r.client.List` and read the response or check if there is any error. 
If there is any error, we're not adding it to `resp.Diagnostics`. 
If I'm not mistaken, we need to `res.Diagnostics.AddError` this error, and then return the method. 

Otherwise, when I have an BA API error, such as `no token` etc., current implementation misses that.

pkg/provider/utils.go
```diff
@@ -35,3 +38,12 @@ func unsupportedWarning(message string) diag.Diagnostics {
 		},
 	}
 }
+
+func fromErr(err error, summary string, args ...any) diag2.Diagnostics {
```
serdardalgic, line 42: This function is quite dangerous. 
When we use `fromErr`, it will remove all the other Diagnostics that were created during any Method (can be warnings etc.) and will return a clean New Error Diagnostic that returns a given error. 

Please use any [Diagnostics](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag#Diagnostics) method to add the error (AddError ideally) instead of completely overriding the errors.

YanniHu1996: Hi Serdar, could you please provide more information about the statement "This function is quite dangerous"? As I understand it, this function returns "Diagnostics," which will be appended to `resp.Diagnostics` using the `Append` method. It should not overwrite any existing diagnostics. The purpose of `fromErr` is to convert errors into `Diagnostics` by accepting an error type parameter. On the other hand, `Diagnostics.AddError` does not accept an error parameter, which means you need to convert the error to a string in your subsequent code

serdardalgic (approved):
The PR looks good to me. Thanks for your work, @YanniHu1996 . 🎉

PR #262, opened 2023-06-11
feat: pgd read

wai-wong-edb:


pkg/provider/data_source_pgd.go
```diff
@@ -0,0 +1,314 @@
...
+	resp.TypeName = req.ProviderTypeName + "_pgd"
+}
+
+// Configure adds the provider configured client to the data source.
+func (d *pgdDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
+	}
+
+	d.client = req.ProviderData.(*api.API)
```
serdardalgic, line 35: After creating a specific client (see my comment in the read method of this data source), can you set that client here? Something like:
```
d.client = req.ProviderData.(*api.API).PgdClient()
```

so that we can call the client functions of the PGDClient within the data source methods.

pkg/provider/data_source_pgd.go
```diff
@@ -0,0 +1,320 @@
...
+					Attributes: map[string]schema.Attribute{
+						"region": schema.StringAttribute{
+							Description: "Witness group region.",
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"id": schema.StringAttribute{
+				Computed: true,
```
serdardalgic, line 223:
```suggestion
				Description: "Datasource ID.",
				Computed: true,
```

serdardalgic (changes requested):
A few minor things, and then, I think this PR is ready to go.

serdardalgic (approved):
Great! Thanks for your hard work on this one, @wai-wong-edb

PR #281, opened 2023-07-06
PGD update

wai-wong-edb:


pkg/provider/data_source_pgd.go
serdardalgic, line 1: It would be helpful to go over the fields of the data source, I guess. 
only project_id and cluster id should suffice to read the PGD cluster as a data source. 
If there is no specific reason, I guess the rest of the fields can be computed: true.

pkg/provider/resource_pgd.go
serdardalgic, line 1: It would be good to go over the description of the fields in the Schema, if there are any discrepancies there.

serdardalgic (commented):
Some minor stuff that caught my attention. 
Let's have a last look on Monday and merge this PR. I think it's in a good shape.

serdardalgic (approved):
I think it's now time to merge this PR :) 👏

PR #287, opened 2023-07-19
fix: fix BaAPIUri fetch for the SDKv2 library

serdardalgic:
We started using mux, and the tfprotov5 server is upgraded to tfprotov6 and we're muxing the SDKv2 library resources of tfprotov5 server and Framework library resources of tfprotov6.

For the provider configuration that uses the Framework library, checking if the data.BaAPIUri is nil or not is enough. But, for the old library SDKv2, those values are set to empty string if they're not present.

We already have this check for BaBearerToken. This commit adds the same check for the BaAPIUri as well.

Once we completely move to the Framework Library, the `== ""` check for both configuration values can be removed.

Please review.


YanniHu1996 (approved):
LGTM

PR #303, opened 2023-08-01
pgd refactor models

wai-wong-edb:
Refactor models to allow unknowns so that we can add extra groups to pgd


serdardalgic (approved):
We have thoroughly tested this branch and looks like it's good to go. 
Thanks a lot for your work @wai-wong-edb , LGTM 👍

PR #310, opened 2023-08-17
ww-pgd-pg-config-defaults

wai-wong-edb:
feat: pg config. Insert default values into plan if it doesn't exist

### DIFF
--- a/pkg/plan_modifier/data_group_custom_diff.go
+++ b/pkg/plan_modifier/data_group_custom_diff.go
@@ -209,6 +209,8 @@ func (m customDataGroupDiffModifier) PlanModifySet(ctx context.Context, req plan
 			}
 
 			// pg config
+			// TODO: We should add the default pg_config values to the plan, so that we show the correct drift.
+			// For details, please check pkg/plan_modifier/pg_config.go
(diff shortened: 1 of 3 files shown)


pkg/plan_modifier/data_group_custom_diff.go
```diff
@@ -209,6 +209,8 @@ func (m customDataGroupDiffModifier) PlanModifySet(ctx context.Context, req plan
 			}
 
 			// pg config
+			// TODO: We should add the default pg_config values to the plan, so that we show the correct drift.
```
serdardalgic, line 212: I'm going to create a ticket for this one.

serdardalgic (approved):
I think it's good to merge for now, there will be a follow-up work for showing the changes in the data_group_custom_diff.go.

pkg/plan_modifier/data_group_custom_diff.go
serdardalgic: I created https://github.com/EnterpriseDB/terraform-provider-biganimal/issues/315 and assigned to you, @wai-wong-edb.

PR #319, opened 2023-08-30
docs: removed pgd upscale not supported note

wai-wong-edb:


serdardalgic (approved):
Yeah, we forgot to remove that one 😅 
Thanks for the PR, @wai-wong-edb, LGTM 👍

PR #325, opened 2023-08-31
feat: maintenance windows for pg clusters

wai-wong-edb:


pkg/provider/resource_cluster.go
```diff
@@ -203,6 +205,35 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+				},
+				Attributes: map[string]schema.Attribute{
+					"is_enabled": schema.BoolAttribute{
+						MarkdownDescription: "Is maintenance window enabled.",
+						Optional:            true,
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseStateForUnknown(),
+						},
+					},
+					"start_day": schema.Float64Attribute{
```
serdardalgic, line 221: Let's add validators for the start_day (and also, maybe start_time).

serdardalgic (changes requested):
One small addition, adding the validators for the CMW fields. 
I'll test it tomorrow and then I can approve. Thank you!

pkg/provider/data_source_cluster.go
```diff
@@ -288,9 +290,32 @@ func (c *clusterDataSource) Schema(ctx context.Context, req datasource.SchemaReq
...
 			},
+			"maintenance_window": schema.SingleNestedAttribute{
+				Computed:            true,
+				MarkdownDescription: "Custom maintenance window.",
+				Attributes: map[string]schema.Attribute{
+					"is_enabled": schema.BoolAttribute{
+						MarkdownDescription: "Is maintenance window enabled.",
+						Computed:            true,
+					},
+					"start_day": schema.Float64Attribute{
```
serdardalgic, line 301: What's the reason that we use Float64Attribute instead of Int64Attribute here? (I checked all the `start_day` implementations, it's `Float64Attribute` everywhere, btw.) 
It's good to have the same implementation everywhere, yet I guess we can use Int64 here instead of Float64. 
WDYT, @wai-wong-edb ?

pkg/provider/data_source_cluster.go
```diff
@@ -288,9 +290,32 @@ func (c *clusterDataSource) Schema(ctx context.Context, req datasource.SchemaReq
...
+						MarkdownDescription: "Start day.",
+						Computed:            true,
+						Validators: []validator.Float64{
+							startDayValidator(),
+						},
```
serdardalgic, line 306: This validator checks only if the value is between 0 and 6. For readability and simplicity, we can directly write it. I would propose something like:
```suggestion
						Validators: []validator.Float64{
							float64validator.Between(0, 6),
						},
```
Directly moving this function here. WDYT, @wai-wong-edb ?

pkg/provider/data_source_cluster.go
```diff
@@ -288,9 +290,32 @@ func (c *clusterDataSource) Schema(ctx context.Context, req datasource.SchemaReq
...
+			"maintenance_window": schema.SingleNestedAttribute{
+				Computed:            true,
+				MarkdownDescription: "Custom maintenance window.",
+				Attributes: map[string]schema.Attribute{
+					"is_enabled": schema.BoolAttribute{
+						MarkdownDescription: "Is maintenance window enabled.",
+						Computed:            true,
+					},
+					"start_day": schema.Float64Attribute{
+						MarkdownDescription: "Start day.",
```
serdardalgic, line 302:
```suggestion
						MarkdownDescription: "The day of week, 0 represents Sunday, 1 is Monday, and so on.",
```
I directly copied from the API description [here](https://portal.biganimal.com/api/docs/#tag/Clusters/operation/createCluster).

If you incorporate this change, please also update in the other recurrences of start day of CMW in biganimal_cluster and biganimal_pgd resource/data-source.

pkg/provider/validators.go
```diff
@@ -60,3 +61,14 @@ func validateUUID(v interface{}, _ cty.Path) diag.Diagnostics {
...
 	return nil
 }
+
+func startDayValidator() validator.Float64 {
+	return float64validator.Between(0, 6)
+}
+
+func startTimeValidator() validator.String {
+	return stringvalidator.RegexMatches(
+		regexp.MustCompile(`^(?:[01]\d|2[0-3]):[0-5]\d$`),
```
serdardalgic, line 71:
```suggestion
		regexp.MustCompile(`^([01]\d|2[0-3]):[0-5]\d$`),
```
We probably don't need the non-capturing group in the regex. It would save minimal space, but I would prefer not to use them to increase the readability of the regex. 

If there is any advantage of using them, and I'm not seeing it, I'd love to learn.

pkg/provider/data_source_cluster.go
```diff
@@ -288,9 +290,32 @@ func (c *clusterDataSource) Schema(ctx context.Context, req datasource.SchemaReq
...
+						Computed:            true,
+						Validators: []validator.Float64{
+							startDayValidator(),
+						},
+					},
+					"start_time": schema.StringAttribute{
+						MarkdownDescription: "Start time.",
```
serdardalgic, line 309:
```suggestion
						MarkdownDescription: "Start time. \"hh:mm\", for example: \"23:59\".",
```
It would be great if you can update this in other recurrences of this start_time.

pkg/provider/resource_cluster.go
```diff
@@ -203,6 +205,41 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
...
 			},
+			"maintenance_window": schema.SingleNestedBlock{
+				MarkdownDescription: "Custom maintenance window.",
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.UseStateForUnknown(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"is_enabled": schema.BoolAttribute{
+						MarkdownDescription: "Is maintenance window enabled.",
+						Optional:            true,
```
serdardalgic, line 216:
```suggestion
						Optional:            true,
						Computed:         true,
```
Otherwise, terraform apply throws the following error when the `maintenance_window` block is not present in the tf file:
```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to biganimal_cluster.single_node_cluster, provider "provider[\"registry.terraform.io/enterprisedb/biganimal\"]" produced an unexpected new value: .maintenance_window: was absent, but now present.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
We need to add `Computed:true` to all the fields in the maintenance_window SingleNestedBlock.

serdardalgic (changes requested):
Thanks for the validators, @wai-wong-edb . I've added some comments about them.
Some improvement suggestions, and some suggestions for fixes. We need to add `Compute: true` otherwise terraform produces errors while writing to the state.

pkg/provider/data_source_cluster.go
wai-wong-edb: Yes this is a previous problem where we are mixing the terraform models with the API models. The terraform models should use int64 and the API models need float64. I can change it to separate the terraform model and API model and the terraform model will convert the int64 to float64. Maybe I could do this for PGD too in this PR as it will affect PGD too

pkg/provider/data_source_cluster.go
wai-wong-edb: sure that is simpler

pkg/provider/validators.go
wai-wong-edb: sure maybe i got the regex a bit wrong

serdardalgic: Nope, your regex is correct. When a group is found, regex captures it in the memory. The non-capturing group doesn't capture it in the memory, just interested in if the regex matches or not. As we're not processing the captured groups, it makes sense to use the non-capturing groups, but it just makes the regex harder to read. 🤷 
That was what I meant in the above comment. 
Other than that, both regex strings work as expected.

pkg/provider/resource_cluster.go
serdardalgic: They're already there.

pkg/provider/validators.go
serdardalgic: https://manifold.net/doc/radian/why_do_non-capture_groups_exist_.htm

PR #333, opened 2023-09-06
feat: Add support for CSP subscription ID and service account IDs for BAH clusters

edbamolvyavahare:


pkg/provider/resource_cluster.go
```diff
@@ -66,6 +66,8 @@ type ClusterResourceModel struct {
 	AllowedIpRanges            []AllowedIpRangesResourceModel     `tfsdk:"allowed_ip_ranges"`
 	CreatedAt                  types.String                       `tfsdk:"created_at"`
 	MaintenanceWindow          *commonTerraform.MaintenanceWindow `tfsdk:"maintenance_window"`
+	ServiceAccountIds          types.List                         `tfsdk:"service_account_ids"`
```
wai-wong-edb, line 69: Better to use types.Set. List have to persist order and the response API does not persist order which makes list difficult to maintain as you may have to sort it in many places

pkg/provider/resource_cluster.go
```diff
@@ -366,6 +368,19 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 					},
 				},
 			},
+			"service_account_ids": schema.ListAttribute{
```
wai-wong-edb, line 371: Better to use SetAttribute

pkg/provider/resource_cluster.go
```diff
@@ -366,6 +368,19 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+				MarkdownDescription: "A Google Cloud Service Account is used for logs. If you leave this blank, then you will be unable to access log details for this cluster. Required when cluster is deployed on BigAnimal's cloud account.",
+				Optional:            true,
+				Computed:            true,
+				ElementType:         types.StringType,
+			},
+
+			"pe_allowed_principal_ids": schema.ListAttribute{
```
wai-wong-edb, line 378: Better to use SetAttribute

pkg/provider/resource_cluster.go
```diff
@@ -597,7 +620,7 @@ func (c *clusterResource) ensureClusterIsHealthy(ctx context.Context, cluster Cl
 		})
 }
 
-func makeClusterForCreate(clusterResource ClusterResourceModel) models.Cluster {
+func (c *clusterResource) makeClusterForCreate(ctx context.Context, clusterResource ClusterResourceModel) models.Cluster {
```
wai-wong-edb, line 623: We can revert back to `func makeClusterForCreate(clusterResource ClusterResourceModel) models.Cluster` and leave it alone
and create a method to wrap around it instead. So it would be something like:
`func (c *clusterResource) makeCluster(ctx context.Context, clusterResource ClusterResourceModel) models {
cluster := makeClusterForCreate(clusterResource)
...
<your bah code>
...
}`

and then we can just use `func (c *clusterResource) makeCluster(ctx context.Context, clusterResource ClusterResourceModel) `

WDYT?

pkg/provider/resource_cluster.go
```diff
@@ -652,11 +675,42 @@ func makeClusterForCreate(clusterResource ClusterResourceModel) models.Cluster {
 		}
 	}
 
+	if strings.Contains(clusterResource.CloudProvider.ValueString(), "bah") {
```
wai-wong-edb, line 678: Have this in another function maybe called `setBahSettingsToClusterAs` it would also be easier to test this function in isolation

pkg/provider/resource_cluster.go
```diff
@@ -652,11 +675,42 @@ func makeClusterForCreate(clusterResource ClusterResourceModel) models.Cluster {
...
+			}
+		}
+	}
+
 	return cluster
 }
 
-func makeClusterFoUpdate(clusterResource ClusterResourceModel) *models.Cluster {
-	cluster := makeClusterForCreate(clusterResource)
+func (c *clusterResource) makeClusterFoUpdate(ctx context.Context, clusterResource ClusterResourceModel) *models.Cluster {
```
wai-wong-edb, line 712: Do similar to this, as explained on `makeClusterForCreate`

wai-wong-edb (changes requested):

pkg/provider/resource_cluster.go
serdardalgic: done.

pkg/provider/resource_cluster.go
serdardalgic: done

pkg/provider/resource_cluster.go
serdardalgic: done

pkg/provider/resource_cluster.go
```diff
@@ -597,7 +622,54 @@ func (c *clusterResource) ensureClusterIsHealthy(ctx context.Context, cluster Cl
...
+	clusterRscCSP := clusterResource.CloudProvider
+	clusterRscPrincipalIds := clusterResource.PeAllowedPrincipalIds
+	clusterRscSvcAcntIds := clusterResource.ServiceAccountIds
+
+	// If there is an existing Principal Account Id for that Region, use that one.
+	pids, _ := c.client.GetPeAllowedPrincipalIds(ctx, clusterResource.ProjectId, clusterRscCSP.ValueString(), clusterResource.Region.ValueString())
+	cluster.PeAllowedPrincipalIds = utils.ToPointer(pids.Data)
+
+	// If there is no existing value, user should provide one
+	if cluster.PeAllowedPrincipalIds == nil {
```
wai-wong-edb, line 645: This should be !clusterRscPrincipalIds.isNull(). This will allow the value to be overwritten with the config value. cluster.PeAllowedPrincipalIds contains the server values and only allows config to be used if nil

wai-wong-edb (changes requested):

pkg/provider/resource_cluster.go
serdardalgic: But at this point, we want to check if the server sent any values, regardless of what's provided by the customer. 
Server values always precede if there is already a AllowedPrincipalId and ServerAccountId.

pkg/provider/resource_cluster.go
serdardalgic: I'm not so sure if we need any change here. 
It's already calling the `makeClusterForCreate` method in the next line. And that makeClusterForCreate method has the BAH specific changes. 
If I'm missing something, can you provide more information, please? Thank you.

pkg/provider/resource_cluster.go
```diff
@@ -379,14 +396,22 @@ func (c *clusterResource) Create(ctx context.Context, req resource.CreateRequest
 		return
 	}
 
-	clusterId, err := c.client.Create(ctx, config.ProjectId, makeClusterForCreate(config))
+	cluster, err := c.makeClusterForCreate(ctx, config)
```
serdardalgic, line 399: A suggestion:
```suggestion
	clusterModel, err := c.makeClusterForCreate(ctx, config)
```
For increasing readability? WDYT?
The same would also apply for the method calls in the Update part.

serdardalgic (commented):
Neat job, thanks @wai-wong-edb .
One suggestion to increase the readability. WDYT?

pkg/provider/resource_cluster.go
wai-wong-edb: sure

pkg/provider/resource_cluster.go
```diff
@@ -597,7 +638,63 @@ func (c *clusterResource) ensureClusterIsHealthy(ctx context.Context, cluster Cl
...
+
+func (c *clusterResource) addBAHFields(ctx context.Context, cluster models.Cluster, clusterResource ClusterResourceModel) (models.Cluster, error) {
+	clusterRscCSP := clusterResource.CloudProvider
+	clusterRscPrincipalIds := clusterResource.PeAllowedPrincipalIds
+	clusterRscSvcAcntIds := clusterResource.ServiceAccountIds
+
+	// If there is an existing Principal Account Id for that Region, use that one.
+	pids, err := c.client.GetPeAllowedPrincipalIds(ctx, clusterResource.ProjectId, clusterRscCSP.ValueString(), clusterResource.Region.ValueString())
+	if err != nil {
+		return cluster, err
```
wai-wong-edb, line 659: This needs to return empty cluster
```suggestion
		return models.Cluster{}, err
```

wai-wong-edb (changes requested):

pkg/provider/data_source_cluster.go
```diff
@@ -72,6 +72,8 @@ type clusterDatasourceModel struct {
 	Region                     types.String                        `tfsdk:"region"`
 	FirstRecoverabilityPointAt types.String                        `tfsdk:"first_recoverability_point_at"`
 	MaintenanceWindow          *terraformCommon.MaintenanceWindow  `tfsdk:"maintenance_window"`
+	ServiceAccountIds          []string                            `tfsdk:"service_account_ids"`
```
wai-wong-edb, line 75: needs to use types.set

pkg/provider/data_source_cluster.go
```diff
@@ -72,6 +72,8 @@ type clusterDatasourceModel struct {
...
+	PeAllowedPrincipalIds      []string                            `tfsdk:"pe_allowed_principal_ids"`
```
wai-wong-edb, line 76: needs to use types.set

wai-wong-edb (changes requested):

wai-wong-edb (approved):

PR #350, opened 2023-09-22
feat: PGD BAH

wai-wong-edb:


pkg/api/pgd_client.go
```diff
@@ -131,6 +135,40 @@ func (c PGDClient) CalculateWitnessGroupParams(ctx context.Context, projectId st
 	return &response.Data, err
 }
 
+func (c PGDClient) GetServiceAccountIds(ctx context.Context, projectID string, cspID string, regionID string) (*models.ServiceAccountIds, error) {
```
edbamolvyavahare, line 138: can we not reuse these functions from the cluster client ?

wai-wong-edb: Yes good point

edbamolvyavahare (approved):
LGTM

PR #353, opened 2023-09-25
feat: Added support for superuser access for cluster resource

edbamolvyavahare:


pkg/provider/resource_cluster.go
```diff
@@ -383,6 +384,11 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 				ElementType:         types.StringType,
 				PlanModifiers:       []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
 			},
+
+			"superuser_access": schema.BoolAttribute{
+				MarkdownDescription: "Is superuser access enabled.",
+				Optional:            true,
```
wai-wong-edb, line 390: This should be a computed field only if we don't set it in the request and expect it in the response

wai-wong-edb (changes requested):

pkg/provider/resource_cluster.go
edbamolvyavahare: this is a boolean value which user can set to enable/disable

edbamolvyavahare: maybe we can update the description to me more specific

wai-wong-edb: I see. In that case we should have both optional: true and computed: true. So we can set it and also expect a value back

wai-wong-edb (approved):
LGTM. Tested with `superuser_access = true` and omitted

pkg/provider/resource_cluster.go
edbamolvyavahare: done

wai-wong-edb (approved):

PR #439, opened 2024-01-09
fix: pgd allowed ip ranges bug

wai-wong-edb:
pgd groups needs allowed IP ranges fix when private networking is true and when allowed IP ranges is []


edbamolvyavahare (approved):
LGTM

Issue #455, opened 2024-02-02
Provider produced inconsistent result after apply - `max_worker_processes`


bryan-bar:
**Describe the bug**
Provider versions `v0.6.1` and `0.7.0` creates the expected resources but terraform fails due to an inconsistent result with `pg_config`.
If the `pg_config` setting `max_worker_processes: 64` is configured, the cluster provisions as expected.

**To Reproduce**
Provision a cluster without `pg_config` settings.

**Expected behavior**
Terraform success exit.

**Terraform and Provider Version:**
```shell
$ terraform version
Terraform v1.3.6
on linux_amd64
+ provider registry.terraform.io/bryan-bar/toolbox v0.2.2
+ provider registry.terraform.io/enterprisedb/biganimal v0.7.0
+ provider registry.terraform.io/hashicorp/aws v5.34.0
+ provider registry.terraform.io/hashicorp/azurerm v3.89.0
+ provider registry.terraform.io/hashicorp/google v5.14.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.25.2
+ provider registry.terraform.io/hashicorp/local v2.4.1
+ provider registry.terraform.io/hashicorp/null v3.2.2
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/hashicorp/time v0.10.0
+ provider registry.terraform.io/hashicorp/tls v4.0.5

Your version of Terraform is out of date! The latest version
is 1.7.2. You can update by downloading from https://www.terraform.io/downloads.html
```

**Affected Resource(s):**
- biganimal_cluster resource

**Additional context**
Error:
```
module.biganimal_us_east_1["mydb2"].biganimal_cluster.instance[0]: Still creating... [6m40s elapsed]
╷
│ Warning: PG config changed
│ 
│   with module.biganimal_us_east_1["mydb2"].biganimal_cluster.instance[0],
│   on modules/biganimal/main.tf line 1, in resource "biganimal_cluster" "instance":
│    1: resource "biganimal_cluster" "instance" {
│ 
│ PG config changed from [] to [{"name":"maintenance_work_mem","value":"(0.15 * (ram - shared_buffers) / autovacuum_max_workers) > 1GB ? 1GB : (0.15 * (ram - shared_buffers) / autovacuum_max_workers)"}
[...]

PR #457, opened 2024-02-05
fix: pg config bug

wai-wong-edb:
I have removed default pg configs in the response so only user custom pg configs show like:
image

before it use to have a lot more elements because of the defaults. This is a more future solution as defaults can be added/removed from the backend in future and cause a mismatch in our plan and response state.


edbamolvyavahare (dismissed):
LGTM

edbamolvyavahare (approved):
LGTM

bryan-bar:
Fixes https://github.com/EnterpriseDB/terraform-provider-biganimal/issues/455

PR #470, opened 2024-02-15
feat: pg cluster pause and resume

wai-wong-edb:


pkg/plan_modifier/phase.go
```diff
@@ -26,25 +30,33 @@ func (m customPhaseForUnknownModifier) MarkdownDescription(_ context.Context) st
...
 // PlanModifyString implements the plan modification logic.
 func (m customPhaseForUnknownModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
-	// Do nothing if there is no state value.
-	if req.StateValue.IsNull() {
-		return
+	// req.StateValue.IsNull() means it is creating
+	// req.StateValue here is from the read
+	if !req.StateValue.IsNull() {
+		if !strings.Contains(req.StateValue.String(), models.PHASE_HEALTHY) && !strings.Contains(req.StateValue.String(), models.PHASE_PAUSED) {
+			resp.Diagnostics.AddError("Cluster not ready for update operations", "Cluster not in healthy state for update operations please wait...")
```
edbamolvyavahare, line 37: we may also check the condition if phase == models.PHASE_PAUSED, and notify user that he/she needs to(may want to) set pause = false in order to make the cluster healthy

pkg/provider/data_source_cluster.go
```diff
@@ -371,6 +371,10 @@ func (c *clusterDataSource) Schema(ctx context.Context, req datasource.SchemaReq
 					},
 				},
 			},
+			"pause": schema.BoolAttribute{
+				Description: "Pause cluster. If true it will put the cluster on pause and set the phase as paused, if false it will resume the cluster and set the phase as healthy",
```
edbamolvyavahare, line 375: we can also add : 
- Pausing a cluster allows you to save on compute costs without losing data or cluster configuration settings. 
- While paused, clusters aren't upgraded or patched, but changes are applied when the cluster resumes. Pausing a high availability or Postgres Distributed cluster shuts down all cluster nodes.

wai-wong-edb: I changed the description on pg and pgd have a look

pkg/plan_modifier/phase.go
wai-wong-edb: I removed this from the phase and put the check here https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/881dd62be9110634381631a0739c55eb0a26aaa6/pkg/provider/resource_cluster.go#L584

I found it is better to do the check on the resource controller layer rather than the plan_modifier layer because if they want to see the plan and not modifier anything but state is not "healthy" or "paused" it will throw an error before the diff appears

PR #549, opened 2024-05-02
feat: beacon analytics create

wai-wong-edb:


pkg/provider/resource_beacon_analytics.go
```diff
@@ -293,6 +282,220 @@ func (bar *beaconAnalyticsResource) Schema(ctx context.Context, req resource.Sch
 }
 
 func (bar *beaconAnalyticsResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
edbamolvyavahare, line 284: add some comments for code in this function

PR #562, opened 2024-07-03
feat: TDE for faraway-replica

wai-wong-edb:


vishalsawale9 (approved):
LGTM. It may be worth adding some code comments in the diff.

examples/resources/biganimal_faraway_replica/azure/resource.tf
```diff
@@ -72,4 +72,12 @@ resource "biganimal_faraway_replica" "faraway_replica" {
...
   #   <example_value> # ex: "9334e5e6-7f47-aE61-5A4F-ee067daeEf4A"
   # ]
+
+  # transparent_data_encryption = {
+  #   key_id = <example_value>
+  # }
+
+  # transparent_data_encryption = {
+  #   key_id = <example_value>
+  # }
```
edbamolvyavahare, line 82: redundant

pkg/provider/resource_fareplica.go
```diff
@@ -277,6 +280,38 @@ func (r *FAReplicaResource) Schema(ctx context.Context, req resource.SchemaReque
...
+						},
+					},
+					"status": schema.StringAttribute{
+						MarkdownDescription: "Status.",
+						Computed:            true,
+					},
+				},
+			},
+			"pg_identity": schema.StringAttribute{
+				MarkdownDescription: "PG Identity.",
```
edbamolvyavahare, line 312: PG Identity required to grant key permissions to activate the cluster.

pkg/provider/resource_fareplica.go
```diff
@@ -478,7 +513,7 @@ func (r *FAReplicaResource) read(ctx context.Context, fAReplicaResourceModel *FA
 	fAReplicaResourceModel.MetricsUrl = apiCluster.MetricsUrl
 	fAReplicaResourceModel.BackupRetentionPeriod = types.StringPointerValue(apiCluster.BackupRetentionPeriod)
 	fAReplicaResourceModel.PrivateNetworking = types.BoolPointerValue(apiCluster.PrivateNetworking)
-
+	fAReplicaResourceModel.PgIdentity = types.StringValue(*apiCluster.PgIdentity)
```
edbamolvyavahare, line 516: this should also be defined under
if apiCluster.EncryptionKey != nil {} block

edbamolvyavahare, line 516: may get a nil error here, as it is only applicable if the cluster is TDE enabled and key is set

pkg/provider/resource_fareplica.go
```diff
@@ -277,6 +280,38 @@ func (r *FAReplicaResource) Schema(ctx context.Context, req resource.SchemaReque
 				ElementType:   types.StringType,
 				PlanModifiers: []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
 			},
+			"transparent_data_encryption": schema.SingleNestedAttribute{
```
edbamolvyavahare, line 283: TDE can only be set for faraway-replica if source cluster is TDE enabled, need to validate the condition or need to rely on API error

examples/resources/biganimal_faraway_replica/azure/resource.tf
wai-wong-edb: Don't they need this?

pkg/provider/resource_fareplica.go
wai-wong-edb: I was thinking to let them set transparent_data_encryption.key_id and they should get an API error if it fails.

examples/resources/biganimal_faraway_replica/azure/resource.tf
edbamolvyavahare, line 82: it is mentioned twice

pkg/provider/resource_fareplica.go
edbamolvyavahare: in that case we need to check if the API error gives enough information or else we should show some graceful error relateable to context

examples/resources/biganimal_faraway_replica/azure/resource.tf
wai-wong-edb: oh right

pkg/provider/resource_fareplica.go
wai-wong-edb: Ok. I did do a test it seems like faraway-replica is broken. I think I fixed it in v1 I recall, so I think I will test it again after v1 is merged rather than trying to fix it in this PR

PR #568, opened 2024-07-31
fix: v1 regression fixes

wai-wong-edb:


pkg/plan_modifier/pg_bouncer.go
```diff
@@ -39,51 +39,64 @@ func (m CustomPgBouncerModifier) PlanModifyObject(ctx context.Context, req planm
...
-				stateSettings = req.StateValue.Attributes()["settings"].(basetypes.SetValue).Elements()
-			}
+	if !req.PlanValue.IsNull() {
+		reqPlanSettings := req.PlanValue.Attributes()["settings"].(basetypes.SetValue)
+		// have plan settings combine with state settings if planned settings len > 0
+		if len(reqPlanSettings.Elements()) > 0 {
+			reqPlanIsEnabled := req.PlanValue.Attributes()["is_enabled"].(basetypes.BoolValue)
+			if !reqPlanIsEnabled.ValueBool() {
+				if !req.PlanValue.Attributes()["settings"].(basetypes.SetValue).IsNull() {
+					resp.Diagnostics.AddError("if pg_bouncer.is_enabled = false then pg_bouncer.settings should be removed", "please remove pg_bouncer.settings if pg_bouncer.is_enabled = false")
```
vishalsawale9, line 49: Would it handle the `settings = []` case?

wai-wong-edb: I think settings = [] is already handled below if enabled is true. It doesn't allow it. But I put in another check so if enabled is false it  shouldn't allow settings = [] also

edbamolvyavahare (approved):
LGTM

PR #574, opened 2024-08-07
feat: Read-only connection string for PGD

vishalsawale9:
This PR intends to support the RO connection string & RO workloads for the PGD cluster through the terraform BA provider.


pkg/provider/resource_pgd.go
```diff
@@ -331,6 +331,11 @@ func PgdSchema(ctx context.Context) schema.Schema {
 								},
 							},
 						},
+						"ro_connection_uri": schema.StringAttribute{
+							MarkdownDescription: "Cluster read-only connection URI. Only available for high availability clusters.",
```
wai-wong-edb, line 335: remove `Only available for high availability clusters`

pkg/provider/resource_pgd.go
```diff
@@ -376,6 +381,11 @@ func PgdSchema(ctx context.Context) schema.Schema {
 							ElementType:   types.StringType,
 							PlanModifiers: []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
 						},
+						"read_only_connections": schema.BoolAttribute{
+							Description: "Read-only connections.",
```
wai-wong-edb, line 385: Change to something like `Is read-only connections enabled.`

wai-wong-edb (commented):
Requested changes

pkg/provider/resource_pgd.go
vishalsawale9: Fixed in the latest commit.

pkg/provider/resource_pgd.go
vishalsawale9: Fixed in the latest commit.

wai-wong-edb (approved):
LGTM

PR #582, opened 2024-08-22
feat: tag assign region

wai-wong-edb:


pkg/provider/data_source_region.go
```diff
@@ -90,6 +90,23 @@ func (r *regionsDataSource) Schema(ctx context.Context, req datasource.SchemaReq
 				Description: "Unique region ID. For example, \"germanywestcentral\" in the Azure cloud provider, \"eu-west-1\" in the AWS cloud provider.",
 				Optional:    true,
 			},
+			"tags": schema.SetNestedAttribute{
+				Description: "Assign existing tags or create tags to assign to this resource",
```
vishalsawale9, line 94:
```suggestion
				Description: "Assign existing tags or create new tags to assign to this resource",
```

pkg/provider/resource_region.go
```diff
@@ -83,6 +86,36 @@ func (r regionResource) Schema(ctx context.Context, req resource.SchemaRequest,
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"tags": schema.SetNestedAttribute{
+				Description: "Assign existing tags or create tags to assign to this resource",
```
vishalsawale9, line 90: Same as [above](https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/582/files#r1728390020)

vishalsawale9 (commented):
Left some minor comments. Rest looks good to me.

pkg/provider/data_source_region.go
edbamolvyavahare, line 94: as this is a data_source description we should not mention assign or create, instead it should mention something like "show tags associated with this resource"

wai-wong-edb: I will change it all in a later PR https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/583 so it is done altogether

edbamolvyavahare (approved):
LGTM

PR #583, opened 2024-08-22
feat: tag analytical cluster

wai-wong-edb:


examples/resources/biganimal_analytics_cluster/aws/resource.tf
```diff
@@ -32,6 +32,16 @@ resource "biganimal_analytics_cluster" "analytics_cluster" {
   project_id   = var.project_id
   pause        = false
 
+  #tags = [
+  #  {
+  #     tag_name  = "wai-tag-3"
```
vishalsawale9, line 37: Better give some other name :)

pkg/provider/data_source_analytics_cluster.go
```diff
@@ -193,6 +193,23 @@ func (r *analyticsClusterDataSource) Schema(ctx context.Context, req datasource.
 					"Pausing a high availability cluster shuts down all cluster nodes",
 				Optional: true,
 			},
+			"tags": schema.SetNestedAttribute{
+				Description: "Assign existing tags or create tags to assign to this resource",
```
vishalsawale9, line 197: https://github.com/EnterpriseDB/terraform-provider-biganimal/pull/582/files#r1728390020

vishalsawale9 (commented):
A few neat comments, but it looks good to me overall.

pkg/provider/data_source_analytics_cluster.go
edbamolvyavahare, line 197: as this is a data_source description, we should not mention assign or create, instead it should mention something like "show tags associated with this resource"

edbamolvyavahare (approved):
LGTM

PR #584, opened 2024-09-05
Docs: Remove the bearer token reference from README docs

vishalsawale9:
Remove the bearer token reference from README docs, as it is a deprecated method for BigAnimal's provider authentication. We are keeping its reference in the code but removing it from the README to discourage users from using it as env var.


docs/index.md
```diff
@@ -22,9 +22,7 @@ see [Getting started with the BigAnimal free trial.](https://www.enterprisedb.co
 ```terraform
 # Configure the BigAnimal Provider
 provider "biganimal" {
-  ba_bearer_token = "<redacted>"
```
wai-wong-edb, line 25: We need to modify the example file too https://github.com/EnterpriseDB/terraform-provider-biganimal/blob/main/examples/provider/provider.tf. Remove bearer token in the example file and also remove the access key comments

wai-wong-edb (changes requested):
1 more thing needs to be changed

docs/index.md
vishalsawale9: @wai-wong-edb : Done. Please check.

PR #588, opened 2024-09-19
chore: Update AWS instance type example to use series 6

vishalsawale9:
This PR is intended to update the cluster provisioning options in AWS biganimal TF provider to use the 6 series instead of the 5 series.


wai-wong-edb (approved):
LGTM

PR #607, opened 2024-11-08
fix: validation check. Prompt to remove pe ids and service ids if clo…

wai-wong-edb:
…ud provider is not bah


edbamolvyavahare (approved):
LGTM

PR #608, opened 2024-11-11
feat: wal storage

wai-wong-edb:


examples/resources/biganimal_cluster/ha/resource.tf
```diff
@@ -68,6 +68,12 @@ resource "biganimal_cluster" "ha_cluster" {
     size              = "4 Gi"
   }
 
+  #  wal_storage = {
+  #    volume_type       = "gp3"
+  #    volume_properties = "gp3"
```
edbamolvyavahare, line 73: should we handle setting volume properties internally since it does not have properties like az
` #    volume_type       = "azurepremiumstorage"
  #    volume_properties = "P1"`

wai-wong-edb: oh for AWS/GCP? Possibly it could work but maybe we can do it in another PR as an enhancement so that they can remove volume_properties for AWS/GCP. As of right now I think I recall in the API(for BA anyways) the volume_properties have to match the volume_type otherwise it will throw an error

pkg/provider/common.go
```diff
@@ -31,3 +34,35 @@ func buildAPIReqAssignTags(tfRsrcTags []commonTerraform.Tag) []commonApi.Tag {
 	}
 	return tags
 }
+
+var resourceWal = schema.SingleNestedAttribute{
+	Description: "Write-Ahead Logs (WAL) Storage.",
+	Optional:    true,
+	Attributes: map[string]schema.Attribute{
```
edbamolvyavahare, line 41: can also add more details to description like:

Use a separate storage volume for Write-Ahead Logs (Recommended for high write workloads)

pkg/provider/common.go
```diff
@@ -31,3 +34,35 @@ func buildAPIReqAssignTags(tfRsrcTags []commonTerraform.Tag) []commonApi.Tag {
...
+			Optional:      true,
+			Computed:      true,
+			PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+		},
+		"volume_properties": schema.StringAttribute{
+			Description: "Volume properties in accordance with the selected volume type.",
+			Required:    true,
+		},
+		"volume_type": schema.StringAttribute{
+			Description: "Volume type. For Azure: \"azurepremiumstorage\" or \"ultradisk\". For AWS: \"gp3\", \"io2\", org s \"io2-block-express\". For Google Cloud: only \"pd-ssd\".",
```
edbamolvyavahare, line 64: typo

pkg/provider/common.go
```diff
@@ -31,3 +34,35 @@ func buildAPIReqAssignTags(tfRsrcTags []commonTerraform.Tag) []commonApi.Tag {
...
+	Attributes: map[string]schema.Attribute{
+		"iops": schema.StringAttribute{
+			Description:   "IOPS for the selected volume. It can be set to different values depending on your volume type and properties.",
+			Optional:      true,
+			Computed:      true,
+			PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+		},
+		"size": schema.StringAttribute{
+			Description:   "Size of the volume. It can be set to different values depending on your volume type and properties.",
+			Required:      true,
```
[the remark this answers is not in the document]
wai-wong-edb: size is actually a mandatory field. In the request if I just put P1 and leave size as nil it will return with this:

> biganimal_cluster.single_node_cluster: Creating...
╷
│ Error: status: 400 - Bad Request
│
│   with biganimal_cluster.single_node_cluster,
│   on resource.tf line 21, in resource "biganimal_cluster" "single_node_cluster":
│   21: resource "biganimal_cluster" "single_node_cluster" {
│
│ must have required property 'clusterType'
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties
│ must NOT have additional properties

however if I put P1 and give a size of 7 Gi I get the error:

> biganimal_cluster.single_node_cluster: Creating...
biganimal_pgd.pgd_cluster: Destruction complete after 5s
╷
│ Error: status: 400 - Bad Request
│
│   with biganimal_cluster.single_node_cluster,
│   on resource.tf line 21, in resource "biganimal_cluster" "single_node_cluster":
│   21: resource "biganimal_cluster" "single_node_cluster" {
│
│ Value of storage size must be between 4 Gi and 4 Gi

I guess a future improvement would be to fetch the list of azurepremiumstorage volume properties and their sizes so size can be optional if they select azurepremiumstorage. But I think we can have that as an enhancement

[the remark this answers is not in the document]
edbamolvyavahare: volume_properties: P1
volume_type: P1
size: 4  (exact size associated with each volume_property)

so user need to fill above details, for it to work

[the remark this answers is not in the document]
edbamolvyavahare: let provide link to volume storage ref in examples

edbamolvyavahare (approved):
LGTM

PR #620, opened 2025-04-22
fix: remove tag color from cluster resources

wai-wong-edb:


pkg/provider/resource_cluster.go
```diff
@@ -549,7 +549,7 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 			"service_name": schema.StringAttribute{
 				MarkdownDescription: "Cluster connection service name.",
 				Computed:            true,
-				PlanModifiers:       []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+				PlanModifiers:       []planmodifier.String{plan_modifier.CustomConnection()},
```
edbamolvyavahare, line 552: should have more generic name

